PR #1, opened 2024-09-05
Mic-5288/Pregnancy model

albrja:
### Implementation of pregnancy model
- *Category*: Implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5288
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/concept_model.html

### Changes and notes
-moves over old pregnancy model code and makes updates for new model
-adds custom clock and time interface 

### Verification and Testing
-Successfully wrote artifact
-Successfully ran simulation with initialized columns and birth outcomes recorded


src/vivarium_gates_mncnh/components/children.py
```diff
@@ -0,0 +1,130 @@
...
+        super().__init__()
+        self.lbwsg = LBWSGDistribution()
+
+    def setup(self, builder: Builder):
+        self.randomness = builder.randomness.get_stream(self.name)
+        self.male_sex_percentage = data_values.INFANT_MALE_PERCENTAGES[
+            builder.data.load(data_keys.POPULATION.LOCATION)
+        ]
+
+    def empty(self, index: pd.Index) -> pd.DataFrame:
```
rmudambi, line 30: Can you give this a better name like `get_child_data_structure()` or something?

src/vivarium_gates_mncnh/components/children.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+
+
+class LBWSGDistribution(Component):
+    def setup(self, builder: Builder):
+        self.randomness = builder.randomness.get_stream(self.name)
+        self.exposure = builder.data.load(data_keys.LBWSG.EXPOSURE).set_index("sex")
+        self.category_intervals = self._get_category_intervals(builder)
+
+    def __call__(self, newborn_sex: pd.Series):
```
rmudambi, line 64: Nit: I think it'd be better to have this be an explicit method called something like `get_exposure()` rather than having the component be a callable itself.

src/vivarium_gates_mncnh/components/children.py
```diff
@@ -0,0 +1,130 @@
...
+            {
+                "sex_of_child": sex_of_child,
+                "birth_weight": lbwsg["birth_weight"],
+                "gestational_age": lbwsg["gestational_age"],
+            },
+            index=index,
+        )
+
+
+class LBWSGDistribution(Component):
```
rmudambi, line 58: It's not possible to use or subclass the `LBWSGDistribution` component defined in VPH?

src/vivarium_gates_mncnh/components/observers.py
```diff
@@ -0,0 +1,70 @@
...
+from vivarium_public_health.results import ResultsStratifier as ResultsStratifier_
+
+from vivarium_gates_mncnh.constants.data_values import PREGNANCY_OUTCOMES
+
+
+class ResultsStratifier(ResultsStratifier_):
+    def setup(self, builder: Builder) -> None:
+        self.age_bins = self.get_age_bins(builder)
+        # self.start_year = builder.configuration.time.start
+        # self.end_year = builder.configuration.time.end.year
```
rmudambi, line 16: Should these be removed?

src/vivarium_gates_mncnh/components/observers.py
```diff
@@ -0,0 +1,70 @@
...
+            ),
+            requires_columns=["pregnancy_outcome"],
+        )
+
+        builder.results.register_stratification(
+            "sex", ["Female", "Male"], requires_columns=["sex"]
+        )
+
+
+class BirthObserver(Observer):
```
rmudambi, line 46: Do we actually need to record the individual birth outcomes to disk, since we will be using them in the same simulation during the next time-step?

src/vivarium_gates_mncnh/components/pregnancy.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def setup(self, builder: Builder):
+        """Performs this component's simulation setup.
+
+        Parameters
+        ----------
+        builder : `engine.Builder`
+            Interface to several simulation tools.
+        """
+        super().setup(builder)
```
rmudambi, line 54: Nit: you don't need this `super()` call. Its super-class is `Component`, whose `setup()` method is empty.

src/vivarium_gates_mncnh/constants/data_keys.py
```diff
@@ -15,7 +15,7 @@ class __Population(NamedTuple):
     AGE_BINS: str = "population.age_bins"
     DEMOGRAPHY: str = "population.demographic_dimensions"
     TMRLE: str = "population.theoretical_minimum_risk_life_expectancy"
-    ACMR: str = "cause.all_causes.cause_specific_mortality_rate"
+    #    BACKGROUND_MORBIDITY: str = "cause.other_causes.disability_weight"
```
rmudambi, line 18: Should this be deleted?

src/vivarium_gates_mncnh/constants/data_values.py
```diff
@@ -32,3 +33,42 @@
...
+class _Durations(NamedTuple):
+    FULL_TERM_DAYS = 40 * 7
+    POSTPARTUM_DAYS = 6 * 7
+    PARTURITION_DAYS = 1 * 7
+    DETECTION_DAYS = 6 * 7
+    PARTIAL_TERM_DAYS = 24 * 7
+    INTERVENTION_DELAY_DAYS = 8 * 7
+
+
+DURATIONS = _Durations()
```
rmudambi, line 57: Are these used? I think many (if not all) of these will not be relevant to this sim.

src/vivarium_gates_mncnh/plugins/time.py
```diff
@@ -0,0 +1,36 @@
...
+        super().setup(builder)
+        # TODO: This means currently we need to configure the simulation to only have 4 steps
+        self.simulation_events = (
+            SIMULATION_EVENT_NAMES.INITIALIZATION,
+            SIMULATION_EVENT_NAMES.PREGNANCY,
+            SIMULATION_EVENT_NAMES.INTRAPARTRUM,
+            SIMULATION_EVENT_NAMES.NEONATAL,
+        )
+        self.step_index = 0
+        self.step_name = self.simulation_events[self.step_index]
```
rmudambi, line 31: This is potentially a little confusing, since technically `INITIALIZATION` is on step -1, not 0. Technically, I think you should set step_index to 0 on setup (or `__init__`), but then subtract 1 from it when we call `step_backward()`.

Then `self.step_name` would be set as follows:
```python
self.step_name = self.simulation_events[self.step_index + 1]
```

None of this is a huge deal, though.

src/vivarium_gates_mncnh/components/observers.py
albrja: I don't think so but it was pretty easy to move it over and we don't have to use it in any runs. I can also remove it.

src/vivarium_gates_mncnh/constants/data_values.py
albrja: Yes they get used to make the pregnancy duration choice.

src/vivarium_gates_mncnh/components/children.py
albrja: I spent a little bit of time trying to get this to work but it seems like it is not worth it. This component only wants to sample the exposures and doesn't need the rest of the overhead from risk distributions.

environment.sh
```diff
@@ -10,22 +10,25 @@ Help()
    echo "Syntax: source environment.sh [-h|t|v]"
    echo "options:"
    echo "h     Print this Help."
-   echo "v     Verbose mode."
    echo "t     Type of conda environment. Either 'simulation' (default) or 'artifact'."
+   echo "n     Make a new conda environment even if one currently exists."
```
rmudambi, line 14: See other PR where I suggested `f`

src/vivarium_gates_mncnh/plugins/time.py
```diff
@@ -28,9 +28,13 @@ def setup(self, builder: Builder) -> None:
             SIMULATION_EVENT_NAMES.NEONATAL,
         )
         self.step_index = 0
-        self.step_name = self.simulation_events[self.step_index]
+        self.step_name = self.simulation_events[self.step_index + 1]
 
     def step_forward(self, index: pd.Index) -> None:
         super().step_forward(index)
-        if self.step_index < len(self.simulation_events) - 1:
+        if self.step_index < len(self.simulation_events) - 2:
```
rmudambi, line 35: Doesn't this mean that we will repeat the neonatal step if the user specified too many time-steps? I think we should error during `setup` if there will be more than 3 time-steps to protect against this.

rmudambi (approved):
You can merge this PR, but please do address my comment in the time manager in a subsequent PR

src/vivarium_gates_mncnh/plugins/time.py
albrja: I actually haven't tested that but I do think we should error.

PR #13, opened 2024-12-04
Albrja/mic 5607/maternal hemorrhage implementation

albrja:
### Implement maternal hemorrhage component
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5607
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021_mncnh/maternal_hemorrhage.html#maternal-hemorrhage

### Changes and notes
-refactor maternal sepsis component to be general maternal disorders component
-add new maternal hemorrhage component

### Verification and Testing
-Successfully ran simulation and correct columns were updated on correct timesteps.

### DIFF
--- a/src/vivarium_gates_mncnh/constants/data_values.py
+++ b/src/vivarium_gates_mncnh/constants/data_values.py
@@ -78,7 +78,8 @@ NUM_DRAWS = 500
 class _SimulationEventNames(NamedTuple):
     PREGNANCY = "pregnancy"
     INTRAPARTRUM = "intrapartum"
-    MATERNAL_SEPSIS = "maternal_sepsis"
+    MATERNAL_SEPSIS = "maternal_sepsis_and_other_maternal_infections"
+    MATERNAL_HEMORRHAGE = "maternal_hemorrhage"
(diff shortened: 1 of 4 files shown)


src/vivarium_gates_mncnh/constants/data_values.py
```diff
@@ -78,7 +78,8 @@ class _Durations(NamedTuple):
 class _SimulationEventNames(NamedTuple):
```
rmudambi, line 78: What is this named tuple used for? It's a bit annoying that we'll need to keep this in sync with the list of simulation events defined in the model spec.

rmudambi (approved):

src/vivarium_gates_mncnh/constants/data_values.py
albrja: I am using in (not in this component) but to define the constants here and then check in components on time step and compare them to the constants in this named tuple. But I do agree it is annoying because I can't import this into the model spec.

PR #16, opened 2024-12-06
Albrja/mic 5609/mortality component

albrja:
### Implement cause of death feature of Mortality component
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5609
- *Research reference*: https://github.com/ihmeuw/vivarium_research/pull/1566/files

### Changes and notes
-create custom mortality component
-determine cause of death based on case fatality rates for maternal disorders for each simulant

### Verification and Testing
Ran simulation and simulants died because of expected causes.


src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -0,0 +1,167 @@
...
+        dead_idx = self.randomness.filter_for_probability(
+            choice_data.index,
+            choice_data["mortality_probability"],
+            "mortality_choice",
+        )
+
+        # Update metadata for simulants that died
+        if not dead_idx.empty:
+            pop.loc[dead_idx, COLUMNS.ALIVE] = "dead"
+
```
albrja, line 111: @rmudambi Im leaving everyone as tracked but not sure if I should be doing that.

albrja:
Sorry for all the silly linting updates. The mortality component is the main purpose of this pr.

src/vivarium_gates_mncnh/components/mortality.py
rmudambi: Yes, that's correct. In general, we don't untrack simulants on death, but we especially want to keep them as tracked in this case, since the mother might die, but the child live, and we need to track that child in the neonatal period.

rmudambi (approved):

PR #23, opened 2025-01-06
Albrja/mic-5750/neonatal-mortality-part-1

albrja:
### Adds basic neonatal mortality component to create ACMR pipeline.
- *Category*: Implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5750
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/index.html#cause-model-decision-graph

### Changes and notes
-Adds component for neonatal mortality

### DIFF
--- a/src/vivarium_gates_mncnh/components/mortality.py
+++ b/src/vivarium_gates_mncnh/components/mortality.py
@@ -8,6 +8,7 @@ from vivarium import Component
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.population import SimulantData
+from vivarium_public_health.utilities import get_lookup_columns
 
 from vivarium_gates_mncnh.constants.data_values import (
     COLUMNS,
@@ -83,7 +84,7 @@ class MaternalDisordersBurden(Component):
         )
         self.population_view.update(pop_update)
 
-    def on_time_step(self, event) -> None:
+    def on_time_step(self, event: Event) -> None:
         if self._sim_step_name() != SIMULATION_EVENT_NAMES.MORTALITY:
             return
 
@@ -163,3 +164,79 @@ class MaternalDisordersBurden(Component):
             )
 
         return simulants
+
+
+class NeonatalMortality(Component):
+    """A component to handle neonatal mortality."""
+
+    ##############
+    # Properties #
+    ##############
+
+    @property
+    def configuration_defaults(self) -> dict[str, Any]:
+        return {
+            self.name: {
+                "data_sources": {
+                    "all_cause_mortality_rate": "cause.all_causes.cause_specific_mortality_rate",
+                    "life_expectancy": "population.theoretical_minimum_risk_life_expectancy",
+                }
+            }
+        }
+
+    @property
+    def columns_created(self) -> list[str]:
+        return [COLUMNS.CHILD_CAUSE_OF_DEATH, COLUMNS.CHILD_YEARS_OF_LIFE_LOST]
+
+    @property
+    def columns_required(self) -> list[str]:
+        return [
+            COLUMNS.CHILD_ALIVE,
+        ]
+
+    def setup(self, builder: Builder) -> None:
+        self._sim_step_name = builder.time.simulation_event_name()
+        self.randomness = builder.randomness.get_stream(self.name)
+        self.location = get_location(builder)
+
+        self.acmr = builder.value.register_rate_producer(
+            "all_cause_mortality_rate",
+            source=self.lookup_tables["all_cause_mortality_rate"],
+            component=self,
+            requires_columns=get_lookup_columns(
+                [self.lookup_tables["all_cause_mortality_rate"]]
+            ),
+        )
+
+    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+        pop_update = pd.DataFrame(
+            {
+                COLUMNS.CHILD_CAUSE_OF_DEATH: "not_dead",
+                COLUMNS.CHILD_YEARS_OF_LIFE_LOST: 0.0,
+            },
+            index=pop_data.index,
+        )
+        self.population_view.update(pop_update)
+
+    def on_time_step(self, event: Event) -> None:
+        if self._sim_step_name() not in [
+            SIMULATION_EVENT_NAMES.EARLY_NEONATAL_MORTALITY,
+            SIMULATION_EVENT_NAMES.LATE_NEONATAL_MORTALITY,
+        ]:
+            return
+
+        pop = self.population_view.get(event.index)
+        alive_children = pop.loc[pop[COLUMNS.CHILD_ALIVE] == "alive"]
+        # TODO: add multiple for early vs late neonatal mortality by updating age of children
+        dead_idx = self.randomness.filter_for_rate(
+            alive_children.index, self.acmr(alive_children.index), "neonatal_mortality_choice"
+        )
+
+        if not dead_idx.empty:
+            pop.loc[dead_idx, COLUMNS.CHILD_ALIVE] = "dead"
+            pop.loc[dead_idx, COLUMNS.CHILD_CAUSE_OF_DEATH] = "other_causes"
+            pop.loc[dead_idx, COLUMNS.CHILD_YEARS_OF_LIFE_LOST] = self.lookup_tables[
+                "life_expectancy"
+            ](dead_idx)
+
+        self.population_view.update(pop)
--- a/src/vivarium_gates_mncnh/components/mortality.py
+++ b/src/vivarium_gates_mncnh/components/mortality.py
@@ -180,7 +180,7 @@ class NeonatalMortality(Component):
         return {
             self.name: {
                 "data_sources": {
-                    "probability_death_by_age_group": self.load_probability_death_by_age_group,
+                    "all_cause_mortality_rate": self.load_acmr,
                     "life_expectancy": self.load_life_expectancy_data,
                 }
             }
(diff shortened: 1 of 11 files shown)


src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +164,79 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+    # Properties #
+    ##############
+
+    @property
+    def configuration_defaults(self) -> dict[str, Any]:
+        return {
+            self.name: {
+                "data_sources": {
+                    "all_cause_mortality_rate": "cause.all_causes.cause_specific_mortality_rate",
+                    "life_expectancy": "population.theoretical_minimum_risk_life_expectancy",
```
albrja, line 182: Both mortality components (for mothers and now children) have this same lookup table but that seems fine since the lookup tables are tied to the components.

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +164,79 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+        return [
+            COLUMNS.CHILD_ALIVE,
+        ]
+
+    def setup(self, builder: Builder) -> None:
+        self._sim_step_name = builder.time.simulation_event_name()
+        self.randomness = builder.randomness.get_stream(self.name)
+        self.location = get_location(builder)
+
+        self.acmr = builder.value.register_rate_producer(
```
albrja, line 202: ACMR is a rate so I assume it should be a rate producer vs a value producer. Is that right?

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +164,79 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+        if self._sim_step_name() not in [
+            SIMULATION_EVENT_NAMES.EARLY_NEONATAL_MORTALITY,
+            SIMULATION_EVENT_NAMES.LATE_NEONATAL_MORTALITY,
+        ]:
+            return
+
+        pop = self.population_view.get(event.index)
+        alive_children = pop.loc[pop[COLUMNS.CHILD_ALIVE] == "alive"]
+        # TODO: add multiple for early vs late neonatal mortality by updating age of children
+        dead_idx = self.randomness.filter_for_rate(
```
albrja, line 231: Similar to above, since ACMR is a rate I'm doing filter for rate vs filter for probability. This seems right but I am not 100% sure.

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +166,120 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+    def load_probability_death_by_age_group(self, builder: Builder) -> pd.DataFrame:
+        """Load all-cause mortality rate data."""
+        acmr = builder.data.load("cause.all_causes.cause_specific_mortality_rate")
+        child_acmr = acmr.rename(columns=CHILD_LOOKUP_COLUMN_MAPPER)
+        # Scale by duration of age group
+        child_acmr["value"] = child_acmr["value"] * (
+            child_acmr["child_age_end"] - child_acmr["child_age_start"]
+        )
+        # Convert rate to probability
+        scaled_probability = rate_to_probability(child_acmr["value"])
```
albrja, line 273: @rmudambi Since we set the default I don't need to the scaling factor here correct?

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +166,120 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+
+    def load_probability_death_by_age_group(self, builder: Builder) -> pd.DataFrame:
+        """Load all-cause mortality rate data."""
+        acmr = builder.data.load("cause.all_causes.cause_specific_mortality_rate")
+        child_acmr = acmr.rename(columns=CHILD_LOOKUP_COLUMN_MAPPER)
+        # Scale by duration of age group
+        child_acmr["value"] = child_acmr["value"] * (
+            child_acmr["child_age_end"] - child_acmr["child_age_start"]
+        )
+        # Convert rate to probability
```
zmbc, line 272: Sorry if I caused any confusion here, but I believe `filter_for_rate` already does this part

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +166,120 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+    ##################
+
+    def load_probability_death_by_age_group(self, builder: Builder) -> pd.DataFrame:
+        """Load all-cause mortality rate data."""
+        acmr = builder.data.load("cause.all_causes.cause_specific_mortality_rate")
+        child_acmr = acmr.rename(columns=CHILD_LOOKUP_COLUMN_MAPPER)
+        # Scale by duration of age group
+        child_acmr["value"] = child_acmr["value"] * (
+            child_acmr["child_age_end"] - child_acmr["child_age_start"]
+        )
```
zmbc, line 271: 👍

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +166,120 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+    ##############
+    # Properties #
+    ##############
+
+    @property
+    def configuration_defaults(self) -> dict[str, Any]:
+        return {
+            self.name: {
+                "data_sources": {
+                    "probability_death_by_age_group": self.load_probability_death_by_age_group,
```
rmudambi, line 183: Nit: I'd call this probability_death_**in**_age_group

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +166,120 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+    #####################
+    # Lifecycle methods #
+    #####################
+
+    def setup(self, builder: Builder) -> None:
+        self._sim_step_name = builder.time.simulation_event_name()
+        self.randomness = builder.randomness.get_stream(self.name)
+        self.location = get_location(builder)
+
+        self.probability_death_by_age_group = builder.value.register_value_producer(
```
rmudambi, line 209: Same here

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +166,120 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+            SIMULATION_EVENT_NAMES.EARLY_NEONATAL_MORTALITY,
+            SIMULATION_EVENT_NAMES.LATE_NEONATAL_MORTALITY,
+        ]:
+            return
+
+        pop = self.population_view.get(event.index)
+        alive_children = pop.loc[pop[COLUMNS.CHILD_ALIVE] == "alive"]
+
+        # Update age of children to get correctlookup values
+        if self._sim_step_name() == SIMULATION_EVENT_NAMES.EARLY_NEONATAL_MORTALITY:
```
rmudambi, line 239: What component creates the `CHILD_AGE` column? That is the component that should be updating their ages.

rmudambi: Also, you are putting the child's age directly on the boundary between the age groups. Do you know which age group the lookup table will evaluate this as? Maybe they should have their ages set to the midpoint of the age bin instead.

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -163,3 +166,120 @@ def get_proportional_case_fatality_rates(self, simulants: pd.DataFrame) -> pd.Da
...
+            pop.loc[alive_children.index, COLUMNS.CHILD_AGE] = 7 / 365.0
+        else:
+            pop.loc[alive_children.index, COLUMNS.CHILD_AGE] = 28 / 365.0
+        self.population_view.update(pop)
+
+        # Determine which neonates die and update metadata
+        dead_idx = self.randomness.filter_for_probability(
+            alive_children.index,
+            self.probability_death_by_age_group(alive_children.index),
+            f"{self._sim_step_name}_neonatal_mortality_choice",
```
rmudambi, line 249: We actually don't need to add the time-step name here (though it won't hurt anything to do so), since we will get different random numbers anyway, due to the fact that it is being called the second time on a different time-step.

src/vivarium_gates_mncnh/components/mortality.py
rmudambi: Since you multiplied the rates by the length of the age-bins in line 270, you don't need to provide a scaling factor (and you can use `rate_to_probability()` directly from `vivarium`. I suggested the modified utility function so that you wouldn't do the multiplication here and it'd happen inside the utility function.

So you'd not do the multiplication in line 270 and this call would look like:
```python
scaled_probability = rate_to_probability(
    rate=child_acmr["value"],
    scaling_factor=child_acmr["child_age_end"] - child_acmr["child_age_start"],
)
```

src/vivarium_gates_mncnh/constants/data_values.py
```diff
@@ -172,3 +178,10 @@ class __Columns(NamedTuple):
     COLUMNS.MATERNAL_HEMORRHAGE,
     COLUMNS.MATERNAL_SEPSIS,
 ]
+
+
+CHILD_LOOKUP_COLUMN_MAPPER = {
```
rmudambi, line 183: This is probably cleaner as a function like this?
```python
def get_child_lookup_column(column: str) -> str:
    return f"child_{column}"
```

src/vivarium_gates_mncnh/utilities.py
```diff
@@ -181,3 +182,16 @@ def get_random_variable(draw: int, seeded_distribution: SeededDistribution) -> f
...
+    rate: Sequence[float] | NumberLike, duration: int | None = None
+) -> NumericArray:
+    # encountered underflow from rate > 30k
+    # for rates greater than 250, exp(-rate) evaluates to 1e-109
+    # beware machine-specific floating point issues
+
+    rate = np.array(rate)
+    rate[rate > 250] = 250.0
+    probability: NumericArray = 1 - np.exp(-rate)
+    return probability
```
rmudambi, line 197: You forgot to do the scaling. This can also be more simply done like this:
```python
from vivarium.framework.utilities import rate_to_probability as rate_to_probability_

def rate_to_probability(
    rate: Sequence[float] | NumberLike, scaling_factor: float = 1.0
) -> NumericArray:
    return rate_to_probability_(rate * scaling_factor)
```

rmudambi (approved):
Looks good!

PR #25, opened 2025-01-16
Albrja/mic 5747/neonatal cause component

albrja:
### Add neonatal cause component
- *Category*: Implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5747
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/preterm_birth.html

### Changes and notes
-adds neonatal cause component
-updates LBWSG paf in artifact

### DIFF
--- a/src/vivarium_gates_mncnh/components/neonatal_causes.py
+++ b/src/vivarium_gates_mncnh/components/neonatal_causes.py
@@ -1,0 +1,83 @@
+import pandas as pd
+from vivarium import Component
+from vivarium.framework.engine import Builder
+from vivarium.framework.event import Event
+from vivarium.framework.resource import Resource
+
+from vivarium_gates_mncnh.constants.data_values import (
+    CHILD_LOOKUP_COLUMN_MAPPER,
+    COLUMNS,
+    NEONATAL_CAUSES,
+    PIPELINES,
+)
+
+
+class NeonatalCause(Component):
+    @property
+    def initialization_requirements(
+        self,
+    ) -> list[str | Resource]:
+        return ["lbwsg_paf", "death_in_age_group_probability", COLUMNS.GESTATIONAL_AGE]
+
+    @property
+    def configuration_defaults(self) -> dict:
+        return {
+            self.name: {
+                "data_sources": {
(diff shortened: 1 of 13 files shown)


src/vivarium_gates_mncnh/components/neonatal_causes.py
```diff
@@ -0,0 +1,88 @@
...
+            ga_less_than_37 = pop[COLUMNS.GESTATIONAL_AGE] < 37
+            normalized_csmr.loc[ga_less_than_37] = 0
+        return normalized_csmr
+
+    def modify_death_in_age_group_probability(
+        self, index: pd.Index, probability_death_in_age_group: pd.Series
+    ) -> pd.Series:
+        csmr_pipeline = self.csmr(index)
+        # ACMR = ACMR - CSMR + CSMR
+        modified_acmr = probability_death_in_age_group - csmr_pipeline + csmr_pipeline
```
aflaxman, line 87: should perhaps the first of these `csmr_pipeline` calls be different?  Like the .raw part of the pipeline or something?

aflaxman: > @aflaxman Should we be subtracting off the raw csmr - the csmr that is directly from the artifact and not modified by RR or Z?

Hmm, no, we should subtract off the csmr after it has been modified by RR and Z, but add back in the csmr that might be modified _further_ by an intervention, such as CPAP for preterm-with-RDS.

zmbc: I alluded to this on Slack -- all adjustments for causal deletion, such as the subtraction in this line, must be performed _in the baseline scenario_. Typically we do this by moving it outside the simulation (since the simulation is scenario-specific). In LSFF I did it by calculating the baseline scenario value in the simulation regardless of the actual scenario we are simulating.

We could use more general tools for this!

albrja: Ok I can subtract off the source of the pipeline and that way if we have interventions that modify the pipeline, then the pipeline will be updated.

PR #29, opened 2025-01-29
Albrja/mic-5806/split-preterm-birth

albrja:
### Split neonatal preterm birth into two causes - with and without rds
- *Category*: Refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5806
- *Research reference*: 

### Changes and notes
-splits neonatal preterm birth into two causes


src/vivarium_gates_mncnh/components/neonatal_causes.py
```diff
@@ -85,4 +88,15 @@ def get_normalized_csmr(self, index: pd.Index) -> pd.Series:
 
         normalized_csmr = super().get_normalized_csmr(index)
         normalized_csmr.loc[ga_greater_than_37] = 0
+        # Weight csmr for preterm birth with rds
+        if self.neonatal_cause == NEONATAL_CAUSES.PRETERM_BIRTH_WITH_RDS:
```
albrja, line 92: If csmr is a rate here is it still ok to multiply it by a probability since we are "weighting" the csmr to get the correct number of cases?

src/vivarium_gates_mncnh/components/mortality.py
```diff
@@ -207,14 +206,20 @@ def setup(self, builder: Builder) -> None:
...
+            NEONATAL_CAUSES.PRETERM_BIRTH_WITH_RDS,
+            NEONATAL_CAUSES.PRETERM_BIRTH_WITHOUT_RDS,
             NEONATAL_CAUSES.NEONATAL_SEPSIS,
             NEONATAL_CAUSES.NEONATAL_ENCEPHALOPATHY,
             "other_causes",
         ]
 
         # Get neonatal csmr pipelines
-        self.preterm_csmr = builder.value.get_value(PIPELINES.NEONATAL_PRETERM_BIRTH)
+        self.with_rds_csmr = builder.value.get_value(
```
rmudambi, line 217: Nit: call these `preterm_with_rds_csmr` and `preterm_without_rds_csmr`

src/vivarium_gates_mncnh/components/neonatal_causes.py
rmudambi: Yes, because the weighted rates will sum to the total rate.
```
rate * weight + rate * (1 - weight) 
    = rate * (weight + 1 - weight)
    = rate
```

src/vivarium_gates_mncnh/components/observers.py
```diff
@@ -296,8 +295,8 @@ def configuration_defaults(self) -> dict[str, Any]:
     def __init__(self):
         super().__init__(
             burden_disorders=[
-                f"{NEONATAL_CAUSES.PRETERM_BIRTH}_with_rds",
-                f"{NEONATAL_CAUSES.PRETERM_BIRTH}_without_rds",
+                NEONATAL_CAUSES.PRETERM_BIRTH_WITH_RDS,
```
rmudambi, line 298: You're creating this list independently here, in the neonatal mortality component, and in the RR observer. I'd suggest storing this list (without "other_causes") somewhere so you only need to update it in one place.

rmudambi (approved):

PR #61, opened 2025-05-08
Albrja/mic-6037/Update birth observer to record counts

albrja:
### Update to birth observer to record counts
- *Category*:  Observer
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6037
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/concept_model.html#id18

### Changes and notes
-update birth observer to record counts instead of recording the whole state table

### DIFF
--- a/src/vivarium_gates_mncnh/components/observers.py
+++ b/src/vivarium_gates_mncnh/components/observers.py
@@ -3,6 +3,7 @@ from functools import partial
 from typing import Any
 
 import pandas as pd
+from layered_config_tree import LayeredConfigTree
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.results import Observer
@@ -52,7 +53,7 @@ class ResultsStratifier(ResultsStratifier_):
                     PREGNANCY_OUTCOMES.PARTIAL_TERM_OUTCOME,
                 ]
             ),
-            requires_columns=["pregnancy_outcome"],
+            requires_columns=[COLUMNS.PREGNANCY_OUTCOME],
         )
 
         builder.results.register_stratification(
@@ -75,7 +76,6 @@ class ResultsStratifier(ResultsStratifier_):
         builder.results.register_stratification(
             "delivery_facility_type",
             self.delivery_facility_types,
-            excluded_categories=[DELIVERY_FACILITY_TYPES.NONE],
             is_vectorized=True,
             requires_columns=[COLUMNS.DELIVERY_FACILITY_TYPE],
         )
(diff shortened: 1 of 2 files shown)


src/vivarium_gates_mncnh/components/observers.py
```diff
@@ -124,38 +124,31 @@ def map_child_age_groups(self, pop: pd.DataFrame) -> pd.Series:
...
+    #                 "include": ["child_sex", "pregnancy_outcome", "delivery_facility_type"],
+    #             },
+    #         },
+    #     }
 
     def setup(self, builder: Builder) -> None:
         self._sim_step_name = builder.time.simulation_event_name()
 
+    def get_configuration(self, builder: Builder) -> dict[str, Any]:
+        return builder.configuration["stratification"][self.get_configuration_name()]
```
rmudambi, line 142: You don't appear to be using this method at all. Can it be removed?

src/vivarium_gates_mncnh/components/observers.py
```diff
@@ -124,38 +124,31 @@ def map_child_age_groups(self, pop: pd.DataFrame) -> pd.Series:
...
 
 class BirthObserver(Observer):
-
-    COL_MAPPING = {
-        COLUMNS.SEX_OF_CHILD: "sex",
-        COLUMNS.BIRTH_WEIGHT_EXPOSURE: "birth_weight",
-        COLUMNS.GESTATIONAL_AGE_EXPOSURE: "gestational_age",
-        COLUMNS.PREGNANCY_OUTCOME: "pregnancy_outcome",
-    }
+    # @property
```
rmudambi, line 127: Remove this rather than commenting it out?

src/vivarium_gates_mncnh/components/observers.py
```diff
@@ -318,15 +311,8 @@ def configuration_defaults(self) -> dict[str, Any]:
...
-                    "exclude": ["age_group"],
-                    "include": [
-                        "child_age_group",
-                        "child_sex",
-                        "cpap_availability",
-                        "antibiotics_availability",
-                        "probiotics_availability",
-                        "delivery_facility_type",
-                    ],
+                    "exclude": [],
```
rmudambi, line 314: This property is not needed at all, as it's inherited from `Observer`. 

Same for the other instances below.

src/vivarium_gates_mncnh/model_specifications/model_spec.yaml
```diff
@@ -81,6 +81,44 @@ configuration:
     stratification:
         default:
             - 'age_group'
```
rmudambi, line 83: You're excluding `age_group` from every observation, so why is it the default? Wouldn't `child_age_group` be a better default? Depending on how you're feeling, `child_sex` and/or `delivery_facility_choice` might make sense to be in the defaults as well.

albrja: I have updated this to not use defaults. Defaults just do not work for all of the observers because of the wide state table.

rmudambi (approved):

PR #87, opened 2025-07-11
Initial draft of code to generate IV iron effects on LBWSG outcomes

alibow:
This PR only includes code for generating the effect of IV iron on gestational age and birth weight. Subsequent updates will include code to generate the effects of IV iron on stillbirth as well as hemoglobin on neonatal sepsis, but I thought I would make this first PR now to get some feedback before we get in any deeper. The main function at the bottom takes an argument of a results directory and a draw so that we can parallelize over draws on the cluster as each draw (for all locations/sexes) takes 10-15 minutes to run at this point. The code creates a new csv for each draw and then I am expecting the artifact building code will concatenate across all of those files to build a key that contains all of the data.

See notebook that walks through these functions here: https://github.com/ihmeuw/vivarium_research_mncnh_portfolio/blob/main/data_prep/hemoglobin_mediation/function_testing.ipynb

### DIFF
--- a/src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_birth_effect_generation.py
+++ b/src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_birth_effect_generation.py
@@ -10,7 +10,7 @@ from vivarium_gates_mncnh.constants import metadata
 """
 This code is intended to read in the effects of hemoglobin on stillbirth, birth weight
 and gestational age outcomes as estimated from the IHME Hemoglobin team and 
-prepare them for use in our simulation. 
+prepare them for use in our simulation.
 
 Specifically, this code:
 - Standardizes the RR exposure levels to be 1,000 equal increments between 40 and 150 g/L
@@ -26,6 +26,11 @@ IV iron and saves these values for ultimate use in our simulation
 
 Note that this code is organized to perform a calculation for all modeled locations
 for a single draw and therefore can be parallelized by draw.
+
+A note on data inputs for this code: data in the {outcome}_bop_rrs.csv files
+were copied from files uploaded to mnt/team/anemia/pub/bop/sim_studies
+by Ditha Nanditha in the summer of 2025. There is an associated email correspondence between
+Ditha, Ali, Abie, Syl, and Nathaniel for reference.
 """
 
 artifact_directory = (
@@ -70,7 +75,9 @@ def convert_rrs_to_gbd_exposure(rrs, exposure_levels):
     for col in rrs.columns:
         if col.startswith("draw_"):
             y = rrs[col].values
-            f = interp1d(x_old, y, kind="linear", bounds_error=False, fill_value="extrapolate")
+            f = interp1d(
+                x_old, y, kind="linear", bounds_error=False, fill_value="extrapolate"
+            )
             rrs_interp[col] = f(x_new)
     return rrs_interp
 
@@ -95,6 +102,9 @@ def transform_and_reorder_rrs(rrs, exposure_levels):
         .rename("value")
         .reset_index()
     )
+    assert (
+        exposure_levels[0] == 40
+    ), "Expected lowest exposure level equal to 40 g/L, but got something else"
     order = rrs.loc[rrs.risk == exposure_levels[0]]
     order = order.sort_values(by="value")
     order["order"] = np.arange(len(order))
@@ -125,8 +135,15 @@ def load_prepped_rrs(outcome):
 
 def get_lbwsg_metadata():
     """Loads metadata (birth weight and gestational age start/end values) for low birth weight and short gestation exposure categories.
-    Note that this function does not return any actual exposure data."""
+    Note that this function does not return any actual exposure data.
+
+    todo: this code could be improved by using the 'risk_factor.low_birth_weight_and_short_gestation.categories'
+    key only (and not the 'risk_factor.low_birth_weight_and_short_gestation.exposure' key) to achieve the purpose
+    of this method.
+    """
     # there are hard-coded location/sex/draw values here, but these are not used in actual data generation
+    # note: the LBWSG exposure categories are the same across locations/sex/draw/ages in GBD, so the
+    # values used here are arbitrary
     art = Artifact(artifact_directory + "ethiopia.hdf")
     art_exposure = art.load(f"risk_factor.low_birth_weight_and_short_gestation.exposure")[
         "draw_0"
@@ -178,13 +195,13 @@ def get_lbwsg_birth_exposure(location):
     ).reset_index()
     art_exposure = get_lbwsg_metadata()
     exp = exp.merge(art_exposure, on="parameter")
+    exp = (
         exp.set_index([x for x in exp.columns if "draw" not in x])
         .rename_axis("draw", axis="columns")
         .stack()
         .rename("exposure")
         .reset_index()
         .rename(columns={"sex_of_child": "sex"})
-        .rename(columns={0: "exposure", "level_8": "draw", "sex_of_child": "sex"})
     )
     exp["location"] = location
     return exp
@@ -250,7 +267,7 @@ def get_lbwsg_shifts(draw):
     exposure_levels = ptb_rrs.exposure.unique()
 
     results = pd.DataFrame()
-    for location in metadata.LOCATIONS:
+    for location in [location.lower() for location in metadata.LOCATIONS]:
         exp = get_lbwsg_birth_exposure(location)
         exp = exp.loc[exp.draw == draw]
         for sex in ["Male", "Female"]:


src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
```diff
@@ -0,0 +1,218 @@
...
+    ga_shifts['value'] = np.where(ga_shifts.exposure > 100, np.nan, ga_shifts['value']) # missing values for exposure levels not eligible for IV iron
+    bw_shifts['value'] = np.where(bw_shifts.exposure > 100, np.nan, bw_shifts['value']) # missing values for exposure levels not eligible for IV iron
+
+    # ensure that we are not running into issues with shifting beyond the valid exposure range
+        # if these tests fail, it means that for some hemoglobin exposure levels, the IV iron effect moves the expousre level beyond levels that exist in this data frame
+        # and therefore will end up reading data specific to another sex/location demographic group
+        # to fix thise, we would need to add exposure levels to these data frames and make some assumotion about how to fill in the data for them
+        # (probably just assume that they have the same data as the highest existing exposure level)
+    assert ptb_exposure_levels[999 - ptb_iv_iron_exposure_increment] > 100, "Invalid exposure range for IV iron shifts"
+    assert lbw_exposure_levels[999 - lbw_iv_iron_exposure_increment] > 100, "Invalid exposure range for IV iron shifts"
```
zmbc, line 200: This approach with `.shift` kind of scares me. We're depending very strongly on the order of the dataframe, etc. How about this?

```suggestion
    ptb_exposure_levels = sorted(ga_shifts.exposure.unique())
    lbw_exposure_levels = sorted(bw_shifts.exposure.unique())

    ptb_exposure_increment = ptb_exposure_levels[1] - ptb_exposure_levels[0] # change in hemoglobin exposure (g/L) per exposure level used in GBD risk curve
    assert np.allclose(ptb_exposure_levels, np.linspace(ptb_exposure_levels[0], ptb_exposure_levels[1], step=ptb_exposure_increment)
    ptb_iv_iron_exposure_increment = (iv_iron_md / ptb_exposure_increment).round(0).astype(int) 
    lbw_exposure_increment = lbw_exposure_levels[1] - lbw_exposure_levels[0] # change in hemoglobin exposure (g/L) per exposure level used in GBD risk curve
    assert np.allclose(lbw_exposure_levels , np.linspace(lbw_exposure_levels[0], lbw_exposure_levels[1], step=lbw_exposure_increment)
    lbw_iv_iron_exposure_increment = (iv_iron_md / lbw_exposure_increment).round(0).astype(int)

    ga_shifts['value'] = (ga_shifts['exposure'] + increment).map(ga_shifts.set_index('exposure')['value']) - ga_shifts['value']
    bw_shifts['value'] = (bw_shifts['exposure'] + increment).map(bw_shifts.set_index('exposure')['value']) - bw_shifts['value']
    # check no missing values for exposure levels eligible for IV iron
    assert ga_shifts.loc[ga_shifts.exposure <= 100, 'value'].notnull().all()
    assert bw_shifts.loc[bw_shifts.exposure <= 100, 'value'].notnull().all()
    # set all exposure levels not eligible for IV iron to be missing
    ga_shifts['value'] = np.where(ga_shifts.exposure > 100, np.nan, ga_shifts['value']) # missing values for exposure levels not eligible for IV iron
    bw_shifts['value'] = np.where(bw_shifts.exposure > 100, np.nan, bw_shifts['value']) # missing values for exposure levels not eligible for IV iron
```

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
```diff
@@ -0,0 +1,218 @@
...
+    exp_tmrel = exp_sub.reset_index()
+    exp_tmrel['frac_preterm'] = np.where(exp_tmrel.ga_end <=37, 1,
+                                        np.where(exp_tmrel.ga_start >=37, 0,
+                                                (37 - exp_tmrel.ga_start)/(exp_tmrel.ga_end - exp_tmrel.ga_start)))
+    exp_tmrel['frac_lbw'] = np.where(exp_tmrel.bw_end <= 2500, 1,
+                                       np.where(exp_tmrel.bw_start >= 2500, 0,
+                                                (2500 - exp_tmrel.bw_start)/(exp_tmrel.bw_end - exp_tmrel.bw_start)))
+    tmrel_preterm_prevalence = (exp_tmrel.exposure * exp_tmrel.frac_preterm).sum()
+    tmrel_lbw_prevalence = (exp_tmrel.exposure * exp_tmrel.frac_lbw).sum()
+    rr_target = rr_data.loc[(rr_data.exposure == exposure_level)][draw].values[0]
```
zmbc, line 122:
```suggestion
    rr_target = rr_data.loc[(rr_data.exposure == exposure_level)][draw].iloc[0]
```

(I recently learned from @NathanielBlairStahn that `values` is deprecated.)

zmbc (commented):
I'm a bit out of my depth here but I was curious so I took a look, hopefully the comments are helpful!

alibow:
> I'm a bit out of my depth here but I was curious so I took a look, hopefully the comments are helpful!

Thanks @zmbc! These are all helpful and I will work on integrating them.

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
alibow: > This approach with `.shift` kind of scares me. We're depending very strongly on the order of the dataframe, etc. How about this?

Fair enough! The reason I took the `.shift` approach is because once we add the IV iron effect on hemoglobin to the hemoglobin exposure values (shifted exposure), then they no longer exactly match the original exposure values (which are not incremented at nice round numbers), which I think is a requirement for this `.map` approach. So shifting by a set number of index values that corresponds to the hemoglobin exposure I want bypasses this issue. AI tells me that there are some ways we could handle this differently including interpolation between continuous values, nearest neighbor matching, and rounding to the nearest "grid point." I haven't actually tried any of these out, but I worry that we could end up with funny merges using anything that relies on rounding (like two shifted values rounding to the same non-shifted value) -- maybe that is an impossibility. 

I will do a little big of investigation to see if any of these work well for this and otherwise I will add in more tests to ensure things are ordered and incremented as they need to be before running `.shift`

zmbc: > So shifting by a set number of index values that corresponds to the hemoglobin exposure I want bypasses this issue.

Ah okay I think I understand this better now. Interpolation would probably be the slickest solution, and the most consistent with what I presume we are doing in the sim.

But it still seems like it should be possible to do this with `map`. And the upside of `map` is that if it doesn't work, you will know -- you'll get an error if anything is not found. So if we have issues due to rounding, we could e.g. round our exposure values to 4 digits after the decimal before matching, or something like that.

I didn't get it quite right above, though.

I think the lines would look like:

```python
# Round to nearest ptb_exposure_increment
ptb_iv_iron_exposure_increment = (iv_iron_md / ptb_exposure_increment).round(0).astype(int) * ptb_exposure_increment
...
ga_shifts['value'] = (ga_shifts['exposure'] + ptb_iv_iron_exposure_increment).map(ga_shifts.set_index('exposure')['value']) - ga_shifts['value']
```

and then if there was some imprecision, we could round like:

```python
# Round to nearest ptb_exposure_increment
ptb_iv_iron_exposure_increment = (iv_iron_md / ptb_exposure_increment).round(0).astype(int) * ptb_exposure_increment
...
ga_shifts['value'] = (ga_shifts['exposure'] + ptb_iv_iron_exposure_increment).round(4).map(ga_shifts.assign(exposure=lambda df: df.exposure.round(4)).set_index('exposure')['value']) - ga_shifts['value']
```

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
```diff
@@ -0,0 +1,218 @@
...
+    exp = exp.merge(art_exposure, on='parameter')
+    exp = (exp.set_index([x for x in exp.columns if 'draw' not in x])
+        .stack().reset_index()
+        .rename(columns={0:'exposure',
+                            'level_8':'draw',
+                            'sex_of_child':'sex'}))
+    exp['location'] = location
+    return exp
+
+def calculate_shift_from_rr(exp_data, rr_data, location, outcome, sex, draw, exposure_level):
```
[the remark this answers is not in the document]
alibow: Here is a notebook that uses these functions! https://github.com/ihmeuw/vivarium_research_mncnh_portfolio/blob/main/data_prep/hemoglobin_mediation/function_testing.ipynb

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
alibow: Ok the latest commit attempts to improve this using .merge because I was having trouble with other approaches. But unfortunately I didn't notice until plotting them here (https://github.com/ihmeuw/vivarium_research_mncnh_portfolio/blob/main/data_prep/hemoglobin_mediation/function_testing.ipynb) that there are intermittent missing values, I think due to rounding issues. I will revisit this when I'm back on Thursday!

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
```diff
@@ -0,0 +1,367 @@
...
+
+def calculate_and_save_lbwsg_shifts(results_directory, draw):
+    assert os.path.exists(
+        results_directory + "lbwsg_shifts/"
+    ), f"Results directory {results_directory + 'lbwsg_shifts/'} does not exist"
+    lbwsg_shifts = get_lbwsg_shifts(draw)
+    lbwsg_shifts.to_csv(results_directory + "lbwsg_shifts/" + draw + ".csv", index=False)
+
+
+def calculate_and_save_iv_iron_lbwsg_shifts(results_directory, draw):
```
zmbc, line 353: This will be a point to get input from engineers: how should this should ultimately work? Should we make this an executable script that we can run with `python hgb_lbwsg_effect_generation.py`? I know you also mentioned including a notebook that will call these functions, so maybe that is enough.

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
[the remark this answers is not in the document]
zmbc: Depending on what the engineers say, it might make sense to have this notebook in this repo rather than the research repo!

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
```diff
@@ -0,0 +1,371 @@
...
+import numpy as np
+import pandas as pd
+import scipy
+from vivarium import Artifact, InteractiveContext
+
+
+def load_bop_rrs(outcome):
+    """Load burden of proof hemoglobin estimates for specified outcome"""
+    # TODO: move these files into model repo?
+    rrs = pd.read_csv(f"/mnt/team/anemia/pub/bop/sim_studies/{outcome}/inner_draws.csv")
```
alibow, line 12: add these to `src/model_repo/data/hemglobin_effects/` but make the file names more informative when we add them

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
```diff
@@ -0,0 +1,371 @@
...
+            for outcome in lbwsg_shifts.outcome.unique():
+                assert np.all(
+                    lbwsg_shifts.loc[
+                        (lbwsg_shifts.location == location)
+                        & (lbwsg_shifts.sex == sex)
+                        & (lbwsg_shifts.outcome == outcome)
+                    ].exposure.values
+                    == exposure_levels
+                ), "Inconsistencies in exposure level values and/or orders"
+    assert np.allclose(
```
albrja, line 317: I would not have an assert in code you are executing at runtime. If you are worried about something happening then you should throw an error since an error and a failed assertion will both kill the program.

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
albrja: The code here is expected to take a long time to run if I recall correctly? It ultimately doesn't matter so the script can either be an executable script or generated via a notebook but make  sure you document how to run it well enough that someone who has never run it before could just sit down, read the documentation, and then generate the data.

I will say for other code that takes a short amount of time to run, it should just be part of the artifact building code in loader.py or called from any of the functions in loader.py.

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
zmbc: There is [plenty of debate about this](https://discuss.python.org/t/stop-ignoring-asserts-when-running-in-optimized-mode/13132/13) but an `assert` in Python can be seen as just syntactic sugar for raising an exception. I feel like making these errors would just make the code longer and harder to read.

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
```diff
@@ -0,0 +1,218 @@
...
+    lbw_rrs.columns = lbw_rrs.columns.droplevel()
+    lbw_rrs = lbw_rrs.reset_index()
+    return lbw_rrs
+
+def get_lbwsg_metadata():
+    """Loads metadata (birth weight and gestational age start/end values) for low birth weight and short gestation exposure categories.
+    Note that this function does not return any actual exposure data."""
+    # there are hard-coded location/sex/draw values here, but these are not used in actual data generation
+
+    # TODO: is there a better way to load this within the repo?
```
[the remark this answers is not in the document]
alibow: I am keeping this as is given our discussion about data provenance in all team, but I moved the artifact directory to the top of the code and made notes about when it will need to be updated.

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
alibow: Ok, keeping as is!

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
alibow: Planning to follow-up on this issue in a subsequent PR, so marking as resolved for now.

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_lbwsg_effect_generation.py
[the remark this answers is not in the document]
alibow: Just to follow-up on this... there is a testing notebook for the code in this PR that is added in a subsequent PR here: https://github.com/ihmeuw/vivarium_gates_mncnh/pull/114

zmbc:
Is the base branch of this PR correct? I'm seeing a lot of diff that is likely not intended.

alibow:
> > Good question, this happened when @stevebachmeier did some Jenkins magic to make the build checks pass, but I don't really understand what went into that or the implications. Steve, could you please confirm if this PR would be safe to merge if/when approved?
> 
> Oh, this is quite a few commits behind main. You need to pull main and manage merge conflicts before you merge this. I'm actually concerned that this will even allow you to merge since it's deviated from main! (or maybe it won't, but we can't tell until you get the appropriate reviews first?)
> 
> lmk if you need help doing this!
> 
> @alibow @zmbc

Huh, confusing to me how/why it diverged, but yeah pulling from main resolved it! Thanks (fyi @zmbc)

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_birth_effect_generation.py
```diff
@@ -0,0 +1,414 @@
...
+    # TODO: read in IV iron effect size/threshold from the repo/artifact here instead of hard-coding it
+    iv_iron_md = 23  # change in hemoglobin exposure associated with IV iron
+    iv_iron_threshold = 100  # maximum hemoglobin exposure level (g/L) eligible for IV iron
+    import warnings
+
+    warnings.warn(
+        "WARNING: using hard coded placeholder effect for IV iron intervention. Needs to be updated to artifact value when ready."
+    )
+
+    data["exposure"] = data["exposure"]
```
hussain-jafari, line 313: What is this doing?

hussain-jafari (approved):

src/vivarium_gates_mncnh/data/hemoglobin_effects/lbw_bop_rrs.csv
zmbc, line 1: Can we say something somewhere about the provenance of this file?

zmbc (approved):

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_birth_effect_generation.py
```diff
@@ -0,0 +1,412 @@
...
+from vivarium_gates_mncnh.constants import metadata
+
+"""
+This code is intended to read in the effects of hemoglobin on stillbirth, birth weight
+and gestational age outcomes as estimated from the IHME Hemoglobin team and 
+prepare them for use in our simulation. 
+
+Specifically, this code:
+- Standardizes the RR exposure levels to be 1,000 equal increments between 40 and 150 g/L
+- Reorders the draws for GA and BW effects in order of magnitude of risk at the 40 g/L hemoglobin
```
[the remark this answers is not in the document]
alibow: It is the lowest hemoglobin exposure (highest risk exposure) included in the risk estimation. Adding this to the notes, thanks

src/vivarium_gates_mncnh/data/hemoglobin_effects/hgb_birth_effect_generation.py
```diff
@@ -0,0 +1,412 @@
...
+
+def convert_rrs_to_gbd_exposure(rrs, exposure_levels):
+    """Using linear interpolation, make a new dataframe with relative risks specific to the exposure levels used in GBD
+    rather than the exposure levels output from the outcome-specific burden of proof models."""
+    from scipy.interpolate import interp1d
+
+    x_new = exposure_levels
+    rrs_interp = pd.DataFrame({"risk": x_new})
+    # Interpolate for each draw column
+    risk = rrs["risk"].values
```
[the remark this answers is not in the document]
alibow: Hmm, I think I disagree? x_new is the GBD exposure levels that will be present in the result of this function whereas the "risk" values (I'm renaming to x_old for clarity) are the non-standard values that come straight from the burden of proof outputs. Does that make sense?

PR #93, opened 2025-07-18
Hjafari/feature/mic 6215 lbwsg rr cap implementation

hussain-jafari:
### Description
- *Category*: implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6215
- *Research reference*: [model 12.0](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/concept_model.html#id24)

### Changes and notes
Update age bin column on time step prepare.
Weight LBWSG prevalence by proportion of living late neonatals for a given LBWSG category.
Add mortality to LBWSG PAF model.
Update LBWSG PAF model spec.

### Verification and Testing
Ran simulation to completion and checked that calculated PAFs looked reasonable.

### DIFF
--- a/src/vivarium_gates_mncnh/components/lbwsg.py
+++ b/src/vivarium_gates_mncnh/components/lbwsg.py
@@ -286,7 +286,7 @@ class LBWSGPAFCalculationExposure(LBWSGRisk):
         """Update the age bins to match the simulants' ages."""
         pop = self.population_view.subview(["child_age", "sex_of_child"]).get(event.index)
         pop["age_bin"] = pd.cut(pop["child_age"], self.age_bins["age_start"])
-        self.population_view.update(pop['age_bin'])
+        self.population_view.update(pop["age_bin"])
 
     ##################################
     # Pipeline sources and modifiers #
@@ -402,7 +402,7 @@ class LBWSGPAFObserver(Component):
         lbwsg_prevalence = self.lbwsg_exposure.rename(
             {"parameter": "lbwsg_category", "value": "prevalence"}, axis=1
         )
-        
+
         # Subset to age group for exposure - have to use np.isclose because age_start is rounded
         lbwsg_prevalence = lbwsg_prevalence[
             np.isclose(self.lbwsg_exposure.child_age_start, age_start, atol=0.001)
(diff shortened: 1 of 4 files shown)


src/vivarium_gates_mncnh/data/lbwsg_paf.yaml
```diff
@@ -27,12 +28,12 @@ configuration:
         end:
             year: 2022
             month: 1
-            day: 2
-        step_size: 3 # Days
+            day: 10
+        step_size: 8 # Days
     population:
-        population_size: 389_992
+        population_size: 194_996
```
hussain-jafari, line 34: We're halving the population because we're putting twice as many simulants in each sex-specific LBWSG category as before. Since we want to keep the number of simulants in each category the same, we're changing the population.

src/vivarium_gates_mncnh/data/loader.py
```diff
@@ -428,19 +428,17 @@ def load_paf_data(
...
+        filename = "calculated_lbwsg_paf_on_cause.all_causes.all_cause_mortality_risk_preterm.parquet"
 
     location_mapper = {
         "Ethiopia": "ethiopia",
         "Nigeria": "nigeria",
         "Pakistan": "pakistan",
     }
 
-    output_dir = paths.PAF_DIR / location_mapper[location]
+    output_dir = paths.PAF_DIR / "temp_outputs" / location_mapper[location]
```
hussain-jafari, line 441: I'm having difficulty writing to the PAF dir so I've made this temporary directory for now. @albrja I'd like to figure out permissions stuff so we can write to the original directory before merging this PR

src/vivarium_gates_mncnh/components/lbwsg.py
```diff
@@ -391,24 +390,41 @@ def setup(self, builder: Builder) -> None:
             requires_columns=["child_alive", "gestational_age_exposure"],
             additional_stratifications=self.config.include,
             excluded_stratifications=self.config.exclude,
-            when="time_step__prepare",
+            when="time_step",
```
albrja, line 393: I would set the priority with a note as to why.

albrja (approved):

src/vivarium_gates_mncnh/data/lbwsg_paf.yaml
```diff
@@ -3,15 +3,16 @@ components:
         components:
             - PAFResultsStratifier()
             - EvenlyDistributedPopulation()
+            - LBWSGMortality()
```
zmbc, line 6: How are we able to include this component in the PAF sim, when it seems like it would itself require a PAF?

hussain-jafari: We only have ACMR here and define the PAF as 0 in the component itself.

zmbc: If we define the PAF to be zero, won't more simulants die in the ENN period than should, and mess up our LNN exposure distribution?

Since early neonates will get LBWSG RRs (>= 1) applied _on top of_ GBD's ACMR, we would end up with an ACMR higher than GBD, no?

PR #99, opened 2025-07-29
fix RR start and intervals and use location-specific PAFs

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6242
- *Research reference*: [model 13.1](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/concept_model.html#id26)

### Changes and notes
Update NonLogLinearRiskEffect to use max RR value for minimum exposure rather than minimum and don't clip RRs to be greater than or equal to 1.

### Verification and Testing
Plotted relative risk data and checked that relative risks stayed flat once exposure went below 40 and dipped below 1 once exposure went above 120.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ if __name__ == "__main__":
         "gbd_mapping>=4.0.0",
         "layered_config_tree",
         "vivarium>=3.4.7",
-        "vivarium_public_health>=4.2.5,<4.3.0",
+        "vivarium_public_health>=4.2.5",
         "click",
         "jinja2",
         "loguru",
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -46,7 +46,7 @@
         "gbd_mapping>=4.0.0",
         "layered_config_tree",
         "vivarium>=3.4.7",
-        "vivarium_public_health>=4.2.5",
+        "vivarium_public_health>=4.2.5,<4.3.0",
```
stevebachmeier, line 49: I assume this means it was more complicated than just importing from the correct location?

albrja: Im not sure why we can't just use the new path instead of pinning below this version

albrja (approved):

PR #105, opened 2025-08-01
Calculate IFA/MMS gestational age shifts

pletale:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/SSCI-2209
- *Research reference*: 

### Changes and notes
 
I copied this notebook over from the [Nutrition Optimization project](https://github.com/ihmeuw/vivarium_research_nutrition_optimization/blob/data_prep/data_prep/antenatal_interventions/Gestational%20age%20shifts.ipynb) and updated LBWSG exposure to GBD 2021 and IFA baseline coverage to the Health Systems team estimates. 

I am not sure this is the right place for me to save this notebook and the CSVs it generates. Please let me know if I should save them somewhere else!


aflaxman (approved):
We should enable reviewnb to make it easier to comment on notebooks during review.

It looks like there are some lines commented out in cell 28 that will prevent this notebook from running all the way through successfully as is.  Can you make sure it runs from start to finish before calling it done?

It might help one of us in the future if you add more notes between calculations in this notebook, to guide the reader through the steps.

alibow:
Just two things:
1. I think we should move the coverage data from Annie's team into this repo... it sounds like people want everything to run the sim in the actual repo
2. Is 2020 the most recent year estimated from Annie's team? Let's use 2023 if they have it!

Otherwise I think this all looks good to go! Thanks Alix!

PR #108, opened 2025-08-07
LNN preterm prevalence calculation

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6266
- *Research reference*: [LNN preterm prevalence modeling strategy](https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/preterm_birth.html#modeling-strategy)

### Changes and notes
Update README to handle this key. 
Pin VPH and apply relevant results updater changes
Define PretermPrevalenceObserver.
Update loader.py to include late neonatal preterm prevalence.  

### Verification and Testing
Rebuilt artifact and ran sim. Checked that preterm prevalence in artifact were the same for early neonatal as model 13.2 artifact, and that there were new late neonatal rows with smaller prevalences than ENN.

### DIFF
--- a/src/vivarium_gates_mncnh/data/loader.py
+++ b/src/vivarium_gates_mncnh/data/loader.py
@@ -388,13 +388,7 @@ def load_paf_data(
     else:
         filename = "calculated_lbwsg_paf_on_cause.all_causes.all_cause_mortality_risk_preterm.parquet"
 
-    location_mapper = {
-        "Ethiopia": "ethiopia",
-        "Nigeria": "nigeria",
-        "Pakistan": "pakistan",
-    }
-
-    output_dir = paths.PAF_DIR / "temp_outputs" / location_mapper[location]
+    output_dir = paths.PAF_DIR / "temp_outputs" / location.lower()
 
     df = pd.read_parquet(output_dir / filename)
     if "input_draw" in df.columns:
@@ -612,12 +606,7 @@ def load_preterm_prevalence(
 
     # get late neonatal prevalence from our PAF simulation
     filename = "calculated_late_neonatal_preterm_prevalence.parquet"
-    location_mapper = {
-        "Ethiopia": "ethiopia",
-        "Nigeria": "nigeria",
-        "Pakistan": "pakistan",
-    }
-    filepath = paths.PRETERM_PREVALENCE_DIR / location_mapper[location] / filename
+    filepath = paths.PRETERM_PREVALENCE_DIR / location.lower() / filename
     data = pd.read_parquet(filepath)
     data = data.drop(["scenario", "random_seed"], axis=1)
     data = data.pivot(index="child_sex", columns="input_draw", values="value")


src/vivarium_gates_mncnh/components/lbwsg.py
```diff
@@ -474,6 +473,107 @@ def calculate_mortality_weights(self, sex: str) -> pd.Series:
...
+            when="time_step",
+            to_observe=self.to_observe,
+        )
+
+    def on_time_step_cleanup(self, event: Event) -> None:
+        """Increment step number at the end of each time step."""
+        self.step_number += 1
+
+    def calculate_preterm_prevalence(self, x: pd.DataFrame) -> float:
+        # weight preterm prevalence by fraction of simulants who survived to late neonatal period
```
hussain-jafari, line 513: This is all the same as the PAF observer until we get to "Get preterm categories".

src/vivarium_gates_mncnh/components/lbwsg.py
```diff
@@ -474,6 +473,107 @@ def calculate_mortality_weights(self, sex: str) -> pd.Series:
...
+        # within a given LBWSG category
+        # this fraction will be 1 at the first time step because no one has died yet, which is
+        # what we want
+        unique_sexes = x["sex_of_child"].unique()
+        if len(unique_sexes) != 1:
+            raise ValueError(
+                "Stratified data contains more than one sex, but this observer (LBWSGPAFObserver) needs sex-stratified data."
```
zmbc, line 520:
```suggestion
                "Stratified data contains more than one sex, but this observer (PretermPrevalenceObserver) needs sex-stratified data."
```

src/vivarium_gates_mncnh/components/lbwsg.py
```diff
@@ -474,6 +473,107 @@ def calculate_mortality_weights(self, sex: str) -> pd.Series:
...
+        pop_data = self.population_view.get(full_index)[
+            ["lbwsg_category", "child_alive", "sex_of_child"]
+        ]
+        pop_data = pop_data.loc[pop_data["sex_of_child"] == sex]
+        weights = (
+            pop_data.groupby(["lbwsg_category", "sex_of_child"])["child_alive"]
+            .agg(proportion_alive=lambda x: (x == "alive").mean())
+            .reset_index()
+        )
+        return weights
```
zmbc, line 567: Is there a way to share this with the PAF observer to avoid repeating the code?

src/vivarium_gates_mncnh/data/loader.py
```diff
@@ -606,7 +606,36 @@ def load_preterm_prevalence(
...
+    enn_data["child_age_start"] = data_values.EARLY_NEONATAL_AGE_START
+    enn_data["child_age_end"] = data_values.LATE_NEONATAL_AGE_START
+
+    # get late neonatal prevalence from our PAF simulation
+    filename = "calculated_late_neonatal_preterm_prevalence.parquet"
+    location_mapper = {
+        "Ethiopia": "ethiopia",
+        "Nigeria": "nigeria",
+        "Pakistan": "pakistan",
+    }
```
zmbc, line 619: Is this necessary? Why not just `location.lower()`?

src/vivarium_gates_mncnh/data/lbwsg_paf_branches.yaml
zmbc, line 1: Looks like this reverts the changes from #102, was that intentional?

src/vivarium_gates_mncnh/components/lbwsg.py
```diff
@@ -474,6 +473,107 @@ def calculate_mortality_weights(self, sex: str) -> pd.Series:
...
+        preterm_prevalence = (
+            preterm_prevalences["mortality_weighted_prevalence"].sum()
+            / lbwsg_prevalence["mortality_weighted_prevalence"].sum()
+        )
+        return preterm_prevalence
+
+    def results_updater(self, old: pd.DataFrame, new: pd.DataFrame) -> pd.DataFrame:
+        return new
+
+    def calculate_mortality_weights(self, sex: str) -> pd.Series:
```
hussain-jafari, line 555: This is the exact same method that appears in the PAF observer, but it makes references to self so it's not immediately clear to me how/whether to avoid duplication. I considered subclassing but that seems like overengineering?

zmbc (approved):
Overall seems good to me, the real test will be V&V though!

src/vivarium_gates_mncnh/components/lbwsg.py
hussain-jafari: Good catch!

src/vivarium_gates_mncnh/components/lbwsg.py
albrja: + 1 this

src/vivarium_gates_mncnh/data/loader.py
hussain-jafari: I copy pasted this from somewhere else and thought the same but decided not to change it for consistency. Now that you've suggested it though, I'm OK with switching to lower() in both places.

src/vivarium_gates_mncnh/components/lbwsg.py
hussain-jafari: See my other comment:

"This is the exact same method that appears in the PAF observer, but it makes references to self so it's not immediately clear to me how/whether to avoid duplication. I considered subclassing but that seems like overengineering?"

ChatGPT also suggested subclassing

src/vivarium_gates_mncnh/components/lbwsg.py
albrja: Im confused as to what you are saying here? If this method is copy/paste from a different class the point would be to subclass so you don't have to do that.

hussain-jafari: It's not clear to me which approach is better. This current approach is very slightly easier to read and understand because both classes are self contained, and while there is code being duplicated it's only 10 lines. The subclassing approach avoids duplication, but is very slightly harder to read.

src/vivarium_gates_mncnh/components/lbwsg.py
rmudambi:
```python

def calculate_preterm_prevalence(self, x):
    helper_func(self, x)


def helper_func(component, x):
    

```

PR #111, opened 2025-08-12
model 14 ANC implementation

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6296
- *Research reference*: [ANC model](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/antenatal_care_module/module_document.html#antenatal-care-attendance-module)

### Changes and notes
Create lookuptables for 3 input variables ANCfirst, ANC1, and ANC4.
Define new state machine - A, B, C, and D states replacing the "attended" boolean - and new transitions.
subsequenr simulation, observer, and test changes.  

### Verification and Testing
Checked in InteractiveContext that we had an anenatal_care_attendance column which contained expected values - none, first_trimester, etc.

### DIFF
--- a/tests/test_anc.py
+++ b/tests/test_anc.py
@@ -41,6 +41,7 @@ def attended_anc_facility_proportion(
     return attended_facility_proportion
 
 
+@pytest.mark.skip(reason="Model 14 ANC updates have obsoleted this test.")
 def test_attended_care_facility_proportions(
     attended_anc_facility_proportion: float,
     population: pd.DataFrame,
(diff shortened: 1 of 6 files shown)


tests/test_anc.py
```diff
@@ -41,6 +41,7 @@ def attended_anc_facility_proportion(
     return attended_facility_proportion
 
 
+@pytest.mark.skip(reason="Model 14 ANC updates have obsoleted this test.")
```
hussain-jafari, line 44: @albrja how important is to update these tests? we can also talk in person at mid-day

albrja: You don't have to update the tests if it is a lot of work. The whole test suite was my attempt to help resolve a lot of the previous v&v issues we had in the model. We could talk more about updating the tests to reflect your implementation but model tests in general are somewhat difficult to maintain.

I think you should avoid skip all tests because at that point you should remove them but I do think it is useful to have tests around that might offer some sort of validation or sanity check on certain components.

albrja (approved):

PR #116, opened 2025-08-15
Fix `simulate run` workflow

zmbc:
### Description
- *Category*: bugfix
- *JIRA issue*: None
- *Research reference*: None

### Changes and notes

Minor changes to get the workflow described in the README, using `simulate run` rather than `psimulate`, working again.

### Verification and Testing

Ran the README steps.


zmbc:
@rmudambi When using `simulate` rather than `psimulate`, there is no scenario or draw columns in the output of the PAF sim, which is why I made the loader change.

rmudambi (approved):

PR #117, opened 2025-08-20
model 15 facility choice updates (without LBWSG updates)

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6287
- *Research reference*: [Correlated propensities](https://vivarium-research.readthedocs.io/en/latest/models/other_models/facility_choice/index.html#correlated-propensities)

### Changes and notes
Generate correlated propensities with Gaussian copula and save as pipelines.
Update ANC so that TreeMachine starts in an initial state that has propensity based transitions. 
Update probabilities for assigning delivery facility types. 

### Verification and Testing
Ran simulation and used breakpoints to check that distribution of facility type matched probabilities


src/vivarium_gates_mncnh/components/propensity.py
```diff
@@ -0,0 +1,57 @@
...
+    `<component_name>.correlated_propensity` in the simulation.
+    These propensities will not be used by any components as a result of creating these
+    pipelines, including the passed components, and it is the responsibility of each component
+    that wants to use these propensities to access them explicitly."""
+
+    ##############
+    # Properties #
+    ##############
+
+    def __init__(self, *component_names: str):
```
rmudambi, line 25: Nit: I think a better way of doing this would be to have no argument to this component, and have a configuration key in the model spec:
```yaml
correlated_propensities:
    - "component_a"
    - "component_b"
    - ...
```

src/vivarium_gates_mncnh/components/tree.py
```diff
@@ -49,3 +61,55 @@ def transition_side_effect(self, index: pd.Index, _event_time: ClockTime) -> Non
         pop = self.population_view.get(index)
         pop[self.update_column] = self.update_value
         self.population_view.update(pop)
+
+
+class ANCInitialState(State):
```
rmudambi, line 66: Wouldn't it make more sense to define these classes in `antenatal_care.py`?

src/vivarium_gates_mncnh/components/tree.py
```diff
@@ -49,3 +61,55 @@ def transition_side_effect(self, index: pd.Index, _event_time: ClockTime) -> Non
...
+                (transition.output_state, np.array(transition.probability(index)))
+                for transition in self.transitions
+            ]
+        )
+        probabilities = np.transpose(probabilities)
+        outputs, probabilities = self._normalize_probabilities(outputs, probabilities)
+        cdf_probabilities = [np.cumsum(probs) for probs in probabilities]
+        propensities = self.propensity(index)
+        # find index of first cumulative probability that is greater than propensity
+        output_indexes = [
```
rmudambi, line 101: Can't you leverage [`_choice()`](https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/randomness/stream.py#L382) to do this rather than writing new logic?

src/vivarium_gates_mncnh/model_specifications/model_spec.yaml
```diff
@@ -157,6 +158,9 @@ configuration:
                 - 'pregnancy_outcome'
                 - 'azithromycin_availability'
                 - 'misoprostol_availability'
+    propensity_ordering:
+        antenatal_care: ['none', 'later_pregnancy_only', 'first_trimester_only', 'first_trimester_and_later_pregnancy']
```
rmudambi, line 162: Is this more readable?
```suggestion
        antenatal_care: 
            - 'none',
            - 'later_pregnancy_only',
            - 'first_trimester_only',
            - 'first_trimester_and_later_pregnancy'
```

tests/test_anc.py
```diff
@@ -15,6 +15,11 @@
 
 from .utilities import get_interactive_context_state
 
+pytest.skip(
+    allow_module_level=True,
+    reason="Model 14 ANC and model 15 facility choice updates have obsoleted these tests.",
+)
```
rmudambi, line 21: If these tests are obsolete (and will remain so), maybe we should remove them rather than skipping them?

tests/test_anc.py
```diff
@@ -55,14 +59,14 @@ def test_attended_care_facility_proportions(
...
 
-@pytest.mark.skip(reason="Model 14 ANC updates have obsoleted this test.")
 def test_received_ultrasound_proportions(
     attended_anc_facility_proportion: float,
     population: pd.DataFrame,
     fuzzy_checker: FuzzyChecker,
 ) -> None:
     location = population[COLUMNS.LOCATION].iloc[0]
     ultrasound_proportion = ANC_RATES.RECEIVED_ULTRASOUND[location]
+    breakpoint()
```
rmudambi, line 69: Remove

tests/test_intrapartum.py
```diff
@@ -15,6 +15,11 @@
 
 from .utilities import get_interactive_context_state
 
+pytest.skip(
```
rmudambi, line 18: Same here?

src/vivarium_gates_mncnh/components/delivery_facility.py
```diff
@@ -50,15 +54,36 @@ def on_time_step(self, event: Event) -> None:
...
+            [assigned_home_preterm, assigned_home_full_term]
+        ).sort_index()
+        pop.loc[assigned_home, COLUMNS.DELIVERY_FACILITY_TYPE] = DELIVERY_FACILITY_TYPES.HOME
+
+        # BEmONC for lower propensities, CEmonC for higher
+        pop.loc[
+            ~assigned_home, COLUMNS.DELIVERY_FACILITY_TYPE
+        ] = DELIVERY_FACILITY_TYPES.CEmONC
+        is_bemonc = (
+            propensity < self.delivery_facility_probabilities[FACILITY_CHOICE.P_BEmONC]
```
rmudambi, line 83: Are we really supposed to be using the same propensity to determine home vs in facility, and to differentiate between different facility levels? Won't all the lower propensities be given to home birth, and therefore we have not enough assigned to BEmONC.

rmudambi (changes requested):

src/vivarium_gates_mncnh/components/delivery_facility.py
hussain-jafari: No - I remember reminding myself to not use the same propensity while writing the home facility portion and then forgot...To explain my check in the description: I did check the distribution of these categories and saw that no one was assigned BEmONC, but I then looked at the lowest propensity and saw that it was indeed greater than the probability of BEmONC. I thought this was acceptable but probably should have thought about it more.

tests/test_anc.py
hussain-jafari: Sounds good to me - I wanted to make sure that we didn't want to update them instead of deleting.

src/vivarium_gates_mncnh/components/antenatal_care.py
```diff
@@ -27,6 +31,9 @@
 )
 from vivarium_gates_mncnh.utilities import get_location
 
+if TYPE_CHECKING:
+    from vivarium.types import DataInput
```
rmudambi, line 35: You don't need to import this in an `if TYPE_CHECKING:` block

src/vivarium_gates_mncnh/components/tree.py
```diff
@@ -1,25 +1,15 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
```
rmudambi, line 3: Don't need this

rmudambi (approved):

src/vivarium_gates_mncnh/components/delivery_facility.py
```diff
@@ -50,15 +54,37 @@ def on_time_step(self, event: Event) -> None:
...
-        pop.loc[birth_idx, COLUMNS.DELIVERY_FACILITY_TYPE] = delivery_facility_type
+        pop = pop[not_partial_term]
+        propensity = self.propensity(pop.index)
+        is_believed_preterm = pop[COLUMNS.STATED_GESTATIONAL_AGE] < 37
+
+        # home births for lower propensities, facility for higher
+        assigned_home_preterm = (
+            propensity[is_believed_preterm]
+            < self.delivery_facility_probabilities[FACILITY_CHOICE.P_HOME_PRETERM]
+        )
```
alibow, line 68: Can you explain what this is doing? I would expect to see something like
`if is_believed_preterm:
    assigned_home_preterm = (propensity[delivery_facility] < self.delivery_facility_probabilities[FACILITY_CHOICE.P_HOME_PRETERM]
else:
    assigned_home_preterm = (propensity[delivery_facility] < self.delivery_facility_probabilities[FACILITY_CHOICE.P_HOME_FULL_TERM]
`
Does that make sense?

PR #120, opened 2025-08-25
ACS exposure

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6323
- *Research reference*: [baseline coverage](https://vivarium-research.readthedocs.io/en/latest/models/intervention_models/intrapartum/acs_intervention.html#baseline-coverage-data)

### Changes and notes
Add ACSAccess component and event which makes ACS available to those who have CPAP and have a predicted gestational age of between 26 to 33 weeks.
Note that this approach might need to be modified in the future if, for example, modifiers act independently on ACS and CPAP. We could explicitly use the same propensity and I'm happy to do so if asked , but I creating a new component to explicitly handle propensity or updating the InterventionAccess component to more generally handle propensities would tae more time. 

### Verification and Testing
Checked before population update that 1) everyone who had ACS also had CPAP and 2) everyone who had CPAP but not ACS had predicted gestational ages NOT between 26 and 33


albrja (approved):

src/vivarium_gates_mncnh/constants/data_values.py
```diff
@@ -102,6 +102,7 @@ class _SimulationEventNames(NamedTuple):
     AZITHROMYCIN_ACCESS = "azithromycin_access"
     MISOPROSTOL_ACCESS = "misoprostol_access"
     CPAP_ACCESS = "cpap_access"
+    ACS_ACCESS = "acs_access"
```
albrja, line 105: Yea you should definitely update this class to be iterable to make your life easier. I should have done that originally.

alibow (approved):

PR #122, opened 2025-08-26
artifact update for ACS PAF 

hussain-jafari:
### Description
- *Category*: artifact
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6326
- *Research reference*: [alternative PAF derivation](https://vivarium-research.readthedocs.io/en/latest/models/intervention_models/neonatal/cpap_intervention.html#calibration-strategy)

### Changes and notes
Update CPAP and ACS PAF. 
Note that the documentation lacks some details that were used in calculation. In particular, the fact that P(BEmONC) = p(facility) * p(BEmONC|facility) (and analogously for CEmONC) and that the probability of CPAP is marginalized over facility choice.

### Verification and Testing
Built artifact for Ethiopia and checked that the structure was the same as for other intervention PAFs.

### DIFF
--- a/src/vivarium_gates_mncnh/constants/data_values.py
+++ b/src/vivarium_gates_mncnh/constants/data_values.py
@@ -320,6 +320,11 @@ DELIVERY_FACILITY_TYPE_PROBABILITIES = {
         FACILITY_CHOICE.P_BEmONC: 0.340528,
     },
 }
+PROPORTION_IN_FACILITY_DELIVERIES = {
+    "Ethiopia": 0.492568,
+    "Nigeria": 0.520097,
+    "Pakistan": 0.771766,
+}
 # Probability each of these facility types has access to CPAP
 CPAP_ACCESS_PROBABILITIES = {
     "Ethiopia": {
--- a/src/vivarium_gates_mncnh/constants/data_values.py
+++ b/src/vivarium_gates_mncnh/constants/data_values.py
@@ -340,6 +340,8 @@ CPAP_ACCESS_PROBABILITIES = {
 }
 CPAP_RELATIVE_RISK_DISTRIBUTION = get_lognorm_from_quantiles(0.53, 0.34, 0.83)
 
+ACS_RELATIVE_RISK_DISTRIBUTION = get_lognorm_from_quantiles(0.84, 0.72, 0.97)
+
 
 ANTIBIOTIC_FACILITY_TYPE_DISTRIBUTION = {
     # NOTE: This is not being used as of model 8.3
--- a/src/vivarium_gates_mncnh/data/loader.py
+++ b/src/vivarium_gates_mncnh/data/loader.py
@@ -104,6 +104,7 @@ def get_data(
         data_keys.NO_CPAP_RISK.RELATIVE_RISK: load_no_cpap_relative_risk,
         data_keys.NO_CPAP_RISK.PAF: load_no_cpap_paf,
         data_keys.NO_ACS_RISK.RELATIVE_RISK: load_no_acs_relative_risk,
+        data_keys.NO_ACS_RISK.PAF: load_no_acs_paf,
         data_keys.NO_ANTIBIOTICS_RISK.P_ANTIBIOTIC_HOME: load_antibiotic_coverage_probability,
         data_keys.NO_ANTIBIOTICS_RISK.P_ANTIBIOTIC_BEMONC: load_antibiotic_coverage_probability,
         data_keys.NO_ANTIBIOTICS_RISK.P_ANTIBIOTIC_CEMONC: load_antibiotic_coverage_probability,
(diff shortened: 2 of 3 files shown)


rmudambi (approved):

src/vivarium_gates_mncnh/constants/data_values.py
```diff
@@ -320,6 +320,11 @@ class __DeliveryFacilityTypes(NamedTuple):
         FACILITY_CHOICE.P_BEmONC: 0.340528,
     },
 }
+PROPORTION_IN_FACILITY_DELIVERIES = {
+    "Ethiopia": 0.492568,
+    "Nigeria": 0.520097,
+    "Pakistan": 0.771766,
+}
```
alibow, line 327: Is this something that would be useful to have as an actual artifact key? It is covariate_id=51 which would make it easier to generalize to new locations if we ever do that.

src/vivarium_gates_mncnh/data/loader.py
```diff
@@ -508,59 +509,29 @@ def load_no_cpap_relative_risk(
...
-    paf_no_cpap = paf_no_cpap.fillna(0.0)
-    return paf_no_cpap
+    # marginalize over facility (no CPAP at home)
+    p_CPAP = (p_CPAP_BEmONC * p_BEmONC) + (p_CPAP_CEmONC * p_CEmONC)
+    p_no_CPAP = 1 - p_CPAP
+
+    population_average_rr = rr_no_CPAP * p_no_CPAP + 1 * p_CPAP
+    paf_no_CPAP = 1 - 1 / population_average_rr
+
+    return paf_no_CPAP
```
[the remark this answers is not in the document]
alibow: Not incorrectly, but the docs have two alternative strategies for calculating the PAF and the record keeping across them has not been well maintained... Hussain and I found it most simple to update to this strategy and I'm planning on bringing up how we should handle this more generally at tomorrow's RT/SE - here's an existing thread related to this issue: https://github.com/ihmeuw/vivarium_research/pull/1764

src/vivarium_gates_mncnh/data/loader.py
```diff
@@ -577,6 +548,35 @@ def load_no_acs_relative_risk(
...
 
 
+def load_no_acs_paf(
+    key: str, location: str, years: Optional[Union[int, str, List[int]]] = None
+) -> float:
+    p_CPAP_BEmONC = get_data(data_keys.NO_CPAP_RISK.P_CPAP_BEMONC, location, years)
+    p_CPAP_CEmONC = get_data(data_keys.NO_CPAP_RISK.P_CPAP_CEMONC, location, years)
+    rr_no_CPAP = get_data(data_keys.NO_CPAP_RISK.RELATIVE_RISK, location, years)
+    rr_no_ACS = get_data(data_keys.NO_ACS_RISK.RELATIVE_RISK, location, years)
+    p_BEmONC_given_facility = data_values.DELIVERY_FACILITY_TYPE_PROBABILITIES[location][
```
alibow, line 558: Not important to the code, but I wonder if we should rename "p_bemonc" to "bemonc_facility_fraction" or something like that so that it's more clear and we don't accidentally use it as the BEMONC rate in the overall population

src/vivarium_gates_mncnh/data/loader.py
```diff
@@ -577,6 +548,35 @@ def load_no_acs_relative_risk(
...
+        data_values.FACILITY_CHOICE.P_BEmONC
+    ]
+    p_CEmONC_given_facility = 1 - p_BEmONC_given_facility
+    p_BEmONC = (
+        data_values.PROPORTION_IN_FACILITY_DELIVERIES[location] * p_BEmONC_given_facility
+    )
+    p_CEmONC = (
+        data_values.PROPORTION_IN_FACILITY_DELIVERIES[location] * p_CEmONC_given_facility
+    )
+    # marginalize over facility (no CPAP at home)
```
alibow, line 568: it looks like we do have a [variable for CPAP coverage at home](https://github.com/ihmeuw/vivarium_gates_mncnh/blob/86f8e36e8f781549cd6671cba9ad33912638dc4a/src/vivarium_gates_mncnh/constants/data_values.py#L328) that is set to zero. Maybe we should just include that in the equation here so that if that variable is ever changed we won't need to remember to update this equation too?

src/vivarium_gates_mncnh/data/loader.py
alibow: Also, do we need to update this how we've defined these parameters in the PAF equations for all of the other interventions that depend on facility type given that we've changed how we have defined these inputs?

src/vivarium_gates_mncnh/constants/data_values.py
alibow: Oh thanks for bringing this up... I believe that we should be using the mean_value of the covariate data for all draws (no draw level variation) because that's what @NathanielBlairStahn used for the input of his delivery facility model

alibow: I just updated this PR to reflect that :) https://github.com/ihmeuw/vivarium_research/pull/1765

PR #124, opened 2025-08-27
Sbachmei/hotfix/dynamically find conda

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-XYZ
- *Research reference*:

### DIFF
--- a/environment.sh
+++ b/environment.sh
@@ -14,9 +14,9 @@ for conda_path in \
   "$HOME/miniconda3" \
   "$HOME/miniconda" \
   "$HOME/anaconda3" \
-  "/ihme/code/$username/miniconda3" \
-  "/ihme/code/$username/miniconda" \
-  "/ihme/code/$username/anaconda3" \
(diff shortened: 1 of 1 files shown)


environment.sh
```diff
@@ -32,14 +42,6 @@ Help()
    echo "l     Install git lfs."
 }
 
-# Define variables
```
stevebachmeier, line 35: just moved this up so I have access to username

rmudambi: I like defining the variables sooner, but I don't see where you use username up above

rmudambi (approved):

stevebachmeier:
@rmudambi Right, I'm dumb. I _initially_ moved the variables up b/c I _was_ using `username`, but with this dynamic approach I don't need it. But I agree I like to define them up front anyway

albrja (approved):

PR #130, opened 2025-09-08
correlated propensities bugfix

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6374

### Changes and notes
Use partial instead of lambda for source of pipeline.

### Verification and Testing
Ran an interactive sim where I confirmed the propensities were the same without the bugfix, and that they were were different and had expected correlations with the bugfix.

### DIFF
--- a/src/vivarium_gates_mncnh/components/propensity.py
+++ b/src/vivarium_gates_mncnh/components/propensity.py
@@ -1,3 +1,4 @@
+from functools import partial
 from itertools import combinations
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/vivarium_gates_mncnh/components/propensity.py
```diff
@@ -30,7 +31,7 @@ def setup(self, builder: Builder):
         for component in self.component_names:
             builder.value.register_value_producer(
                 f"{component}.correlated_propensity",
-                source=lambda index: self.propensities.loc[index, component],
+                source=partial(self.get_component_propensity, component=component),
```
rmudambi, line 34: That's a sneaky one.

PR #133, opened 2025-09-09
refactor: clarify role of sex in LBWSG PAF calculation exposure

zmbc:
### Description
- *Category*: refactor
- *JIRA issue*: None
- *Research reference*: https://github.com/ihmeuw/vivarium_research/pull/1774

### Changes and notes

Cosmetic changes only to explain the "magic number."

### Verification and Testing

Ran sim.

### DIFF
--- a/src/vivarium_gates_mncnh/components/lbwsg.py
+++ b/src/vivarium_gates_mncnh/components/lbwsg.py
@@ -359,16 +359,19 @@ class LBWSGPAFCalculationExposure(LBWSGRisk):
         pop["age_bin"] = pd.cut(pop["child_age"], self.age_bins["age_start"])
         pop = pop.sort_values(["sex_of_child", "child_age"])
 
+        num_sexes = pop.sex_of_child.nunique()
+        assert num_sexes == 2
+
         lbwsg_categories = self.lbwsg_categories.keys()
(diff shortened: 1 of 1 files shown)


src/vivarium_gates_mncnh/components/lbwsg.py
```diff
@@ -359,16 +359,19 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
         pop["age_bin"] = pd.cut(pop["child_age"], self.age_bins["age_start"])
         pop = pop.sort_values(["sex_of_child", "child_age"])
 
+        num_sexes = pop.sex_of_child.nunique()
+        assert num_sexes == 2
```
albrja, line 363: I really dislike assertions in production code (although one could argue this is really artifact generation code and I would agree with them). Maybe raise a ValueError instead.

albrja (approved):

src/vivarium_gates_mncnh/components/lbwsg.py
hussain-jafari: We talked a bit about this in stand up today. It seems like the only thing we really lose with an assert here is the type error as long as we add an informative message explaining the error.

hussain-jafari (approved):

PR #141, opened 2025-09-15
Calculate low hemoglobin PAFs for maternal disorders

pletale:
### Description

- *Category*: data artifact 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/SSCI-2371
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/risk_effects/hemoglobin/index.html#maternal-disorders 

### Changes and notes
Include code to calculate hemoglobin PAFs on:
- maternal sepsis
- maternal hemorrhage
- depressive disorders
- neonatal sepsis 

We will wait until the GBD 2023 update has begun to actually calculate these values.

Notes on things that will need to be updated in the research docs
- We are not considering correlation between PREG OUTCOME->ANC->IFA->HGB with risk of maternal disorder outcome and induced correlation with other interventions that affect these outcome rates (like misoprostol and hemorrhage) induced through our facility choice model... but perhaps we could if we take more simulation timesteps and we wait until baseline IFA model implementation is resolved
- We are calculating the PAF for PPD based on hemoglobin exposure in pregnancy rather than among those who survive labor. This is a limitation that is hopefully small due to maternal mortality being a rare event. But we can revisit if our model is calibrating poorly. Additionally, we may want to calculate the PAF according to global rather than modeled location risk exposures (as reflected in this ticket https://jira.ihme.washington.edu/browse/SSCI-2415). However, as the computational expense of calculating the PAF of this additional cause is low, I will include it here regardless and we can always perform a separate calculation later if desired, especially given that we have an outstanding decision on whether we will update this data source, described here:https://jira.ihme.washington.edu/browse/SSCI-2422
- We are not calculating the PAF for hemoglobin on neonatal sepsis among the parental hemoglobin exposure distribution specific to neonates who survive the early neonatal time period (but perhaps we could if we take more simulation timesteps)
- We have used 250 draws for the RRs on maternal disorders (rather than 100)

### Verification and Testing
See function tester notebook for check that RR values assigned in this code is the same as those assigned within the simulation. Also see how the calculated PAF values vary with differing population sizes used in calculation.


src/vivarium_gates_mncnh/data/hgb_pafs_maternal_disorders/hgb_maternal_disorder_paf_generation.py
```diff
@@ -0,0 +1,117 @@
...
+    custom_model_specification = build_model_specification(path)
+    del custom_model_specification.configuration.observers
+    custom_model_specification.configuration.input_data.artifact_path = (
+        artifact_directory + location + ".hdf"
+    )
+    draw_num = custom_model_specification.configuration.input_data.input_draw_number
+    draw = "draw_" + str(draw_num)
+    # NOTE: setting population size to what we are using in the simulation for a single draw
+    gbd_draw = "draw_" + str(draw_num % 100)
+    # NOTE: We use only the first 100 draws from GBD, repeating them for later draws.
```
alibow, line 36: We're not using gbd_draw anywhere in this code so these lines can be deleted

src/vivarium_gates_mncnh/data/hgb_pafs_maternal_disorders/hgb_maternal_disorder_paf_generation.py
```diff
@@ -0,0 +1,117 @@
...
+        elif 35 <= age < 40:
+            return "35_to_39"
+        elif 40 <= age < 45:
+            return "40_to_44"
+        elif 45 <= age < 50:
+            return "45_to_49"
+        else:
+            return "50_to_54"
+
+    df["age_group"] = df["age"].apply(assign_gbd_age_group)
```
zmbc, line 92:
```suggestion
    age_groups = pd.cut(df["age"], range(10, 56, 5))
    df["age_group"] = age_groups.left.astype(str) + '_to_' + (age_groups.right - 1).astype(str)
```

src/vivarium_gates_mncnh/data/hgb_pafs_maternal_disorders/function_tester.ipynb
zmbc, line 1: Would be good to show the outputs of the helper functions as well

PR #317, opened 2026-05-15
Neonatal mortality V&V

zmbc:
### Description
- *Category*: test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/SSCI-2352
- *Research reference*: None

### Verification and Testing

<--
*** REMINDER ***
CI WILL NOT RUN ANY TESTS.
-->
- [x] model results directory is up to date

### Next steps

- [x] Merge this pull request
- [ ] THIS IS A SENSITIVITY ANALYSIS/EXPERIMENT -- DO NOT MERGE!


zmbc:
Best commit to look at for the changes from previous model: https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/commit/1a0aecfc44652e91a6d5f5e42b968afed48f9a38/

zmbc:
Notes:
- It seems like there is some new LBWSG miscalibration (between main sim and PAF sim) going on, including in the preterm prevalence
- It seems like the probiotics effect on neonatal sepsis is now knocking it out of calibration (possibly because we never re-ran the sepsis effects data generation?)
- The CPAP and ACS effect on preterm with RDS is now severely miscalibrated? I don't understand how the results V&V is not showing this

zmbc:
@hussain-jafari I'm not considering this "done" per se, but I am maybe a bit stuck and could use some help 😅 I have narrowed it down to a few specific issues:

- It appears that the LBWSG exposure distribution (among live births) is substantially different between the PAF sim and the main sim. I think this could in principle be due to our intervention LBWSG effects but those were present in model 19.1.1 IIUC and the scale of the miscalibration was much, much smaller. So I'm not sure what is happening/what changed.
- My first idea for the second issue noted above is to rerun the sepsis effects generation code, which we need to do anyway. Not sure whether this is a you task or a Tyler task as he comes back to the project.
- There is some weird inconsistency I don't understand between how I am checking preterm-with-RDS CSMRisk in this notebook and in the results V&V. This notebook (interactive sim) claims to show substantial miscalibration. Or maybe this is a difference between the version of code I am running interactively and the version we actually ran with `psimulate`?

tylerdy (approved):
I looked through all this and feel more up to speed on it now. I probably dont have any new thoughts but happy to be someone to talk through these issues with. Seems like the results make sense based the issues you're noting. One thing I was wondering about is what the mutator is, like in the CSMR after mutator plots, which look off but maybe are covered in something you've already pointed out

zmbc:
The mutators are typically the effect of interventions, or in the case of ACMR, the effect of the CSMRs. Something I missed in the Vivarium 4.0 changeover was to print out what the mutators actually are:

image

You can see on the left that the first mutator is the effect of preterm-with-RDS CSMR on ACMR. The first mutator on the right is presumably the same thing 😅 I will print out the names in the next round of V&V!

zmbc:
Link to the diff in the image above: https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/commit/1a0aecfc44652e91a6d5f5e42b968afed48f9a38/?path=d483acb1&cell_index=50&sha=ab1d038bb1b2a05ec45805952dccfb7836eb02ad

PR #319, opened 2026-05-19
Hjafari/feature/mic 7077 aph pph hemoglobin shifts

hussain-jafari:
### Description
- *Category*: implementation 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7077

### Changes and notes
Adds postpartum hemoglobin shift logic for antepartum hemorrhage (APH) and postpartum hemorrhage (PPH) cases.

- At the 0-6 week postpartum event (early_neonatal_mortality): hemorrhage shifts are applied to existing pregnancy hemoglobin for survived simulants with APH/PPH.
- At the 6 week-9 month event (postpartum_hemoglobin_nine_month): pregnancy hemoglobin is replaced with a draw from the non-pregnant ensemble distribution, then hemorrhage shifts are applied. Simulants with both APH and PPH receive both shifts additively.
- Simplified load_hemorrhage_hemoglobin_shift in loader.py to return a single scalar row (no demographic index needed since shifts are age-agnostic).

### Verification and Testing
Ran tests.

<--
*** REMINDER ***
CI WILL NOT RUN ANY TESTS.
-->

### Next steps

- [X] Merge this pull request


hussain-jafari:
Now I have all the context needed. Let me synthesize the findings into a unified review.

## PR Review: APH/PPH Hemoglobin Shifts (MIC-7077)

### Summary
This draft PR adds postpartum hemoglobin shift logic for antepartum hemorrhage (APH) and postpartum hemorrhage (PPH) cases. At the 0-6 week postpartum event (`early_neonatal_mortality`), hemorrhage shifts are applied to the existing pregnancy hemoglobin. At the 6 week-9 month event (`postpartum_hemoglobin_nine_month`), pregnancy hemoglobin is replaced with a draw from the non-pregnant ensemble distribution and hemorrhage shifts are then applied. The PR also simplifies the loader to return a single scalar row, adds utility functions for quantile clipping and distribution parameter extraction, and adds the new simulation event to the model spec.

### Design

1. **`clip_quantiles` mutates its input in-place** — utilities.py: The function modifies the passed Series via `q[q > ...] = ...` and also returns it. This hybrid in-place/functional API invites bugs if callers retain a reference to the original. Use `q.clip(lower=Q_LOWER_BOUND, upper=Q_UPPER_BOUND)` instead, which returns a new Series without mutation and communicates intent more clearly.

2. **Implicit index alignment in `_apply_6w_9m_shifts`** — hemoglobin.py: `hgb` is indexed by `survived_idx` (a subset), but `pph_mask`/`aph_mask` are boolean Series over the full population index. The `hgb.loc[pph_mask]` assignment relies on pandas silently aligning indices. This is correct but fragile and non-obvious. Compute masks against the survived subset directly:
   ```python
   survived_pop = pop.loc[survived_mask]
   pph_mask = survived_pop[COLUMNS.POSTPARTUM_HEMORRHAGE] == True
   hgb.loc[pph_mask] += self.pph_shift_6w_9m
   ```

3. **`== True` comparison on boolean columns** — hemoglobin.py: `pop[COLUMNS.POSTPARTUM_HEMORRHAGE] == True` is non-idiomatic for boolean columns. If the column is guaranteed boolean, use the column directly. If it may contain NaN, use `.fillna(False)` to make intent explicit.

4. **Four flat scalar attributes vs. structured lookup** — hemoglobin.py: Four separately-named `self.*_shift_*` attributes encode a 2×2 matrix (cause × period). This is fine at current scale but would scale poorly if more causes or periods are added. A dict keyed by `(cause_column, period)` would allow the shift application to be a loop rather than explicit branches.

### Maintainability

1. **`on_time_step` handles three events with different semantics undocumented** — hemoglobin.py: `LATER_PREGNANCY_INTERVENTION` captures pregnancy hemoglobin, while `EARLY_NEONATAL_MORTALITY` and `POSTPARTUM_HEMOGLOBIN_NINE_MONTH` capture post-shift values. This asymmetry is implicit. A brief comment distinguishing the pre-partum snapshot from postpartum snapshots would help readers.

2. **NaN replacement with 0 in `_sample_non_pregnant_hemoglobin` is silent** — hemoglobin.py: `result[result.isnull()] = 0` replaces NaN with a physiologically impossible hemoglobin value. No logging indicates when this fallback occurs. Consider adding a comment explaining when NaN is expected, or a `logger.warning` if any are replaced.

3. **`Q_LOWER_BOUND = 0.0011` and `Q_UPPER_BOUND = 0.998` are unexplained** — utilities.py: A comment explaining the rationale (e.g., "matches vivarium_public_health bounds to avoid NaN from ensemble distribution ppf at extremes") would prevent someone from "correcting" these.

4. **Loader uses same shift curve for both APH and PPH without noting it** — loader.py: The `key` selects the day range but both hemorrhage types use the same `pred_data.csv`. This is a domain assumption worth documenting.

### DRY

1. **Duplicated population fetch + survived mask computation** — hemoglobin.py and hemoglobin.py: Both `_apply_0_6w_shifts` and `_apply_6w_9m_shifts` fetch identical columns and compute the same `survived_mask`. Extract a helper:
   ```python
   def _get_postpartum_pop_and_mask(self, index):
       pop = self.population_view.get(index, [...])
       survived_mask = pop[COLUMNS.PREGNANCY_OUTCOME].isin([...])
       return pop, survived_mask
   ```

2. **Duplicated hemorrhage shift application pattern** — The "mask + add shift" pattern appears 4 times (PPH/APH × 2 methods). Extract a helper `_apply_hemorrhage_shifts(hgb, pop, survived_mask, pph_shift, aph_shift)` to consolidate.

### Tests

1. **No automated tests for any new functionality.** The PR adds ~200 lines of logic verified only by a Jupyter notebook that is not collected by pytest. Core methods (`_apply_0_6w_shifts`, `_apply_6w_9m_shifts`, `_sample_non_pregnant_hemoglobin`) and utilities (`clip_quantiles`, `get_risk_distribution_parameter`) all lack test coverage. This is acceptable for a sensitivity analysis/experiment but should be addressed before any merge.

2. **No test for simultaneous PPH + APH** — The logic applies both shifts additively for simulants with both conditions. If refactored to `elif` semantics, this behavior would silently break without a test guarding it.

3. **No test for negative hemoglobin clipping** — Both methods clip to `lower=0`, but no test verifies that a low hemoglobin + large negative shift doesn't produce negative values downstream.

### Documentation

1. **`clip_quantiles` and `get_risk_distribution_parameter` need full NumPy-style docstrings** — utilities.py: Both are public functions (no leading underscore, imported externally). Per team style they should have full docstrings with Parameters/Returns sections.

2. **`load_hemorrhage_hemoglobin_shift` missing Parameters/Returns section** — loader.py: This is a public registered loader function. The docstring omits parameter documentation.

3. **No CHANGELOG entry** — Behavioral changes like postpartum hemoglobin shifts warrant a changelog note, though this may be deferred since it's marked DO NOT MERGE.

### Functionality

1. **Pipeline modifier runs on every pipeline query, not just `on_time_step`** — `_modify_exposure_for_postpartum` is registered as an attribute modifier, meaning *any* component querying the hemoglobin exposure pipeline during `EARLY_NEONATAL_MORTALITY` or `POSTPARTUM_HEMOGLOBIN_NINE_MONTH` will receive the shifted value. This appears intentional (the pipeline represents the "true" hemoglobin at that event), but worth confirming since other components (e.g., anemia screening downstream) would see the shifted value too.

2. **Non-pregnant hemoglobin sampling at 6w-9m occurs per pipeline call, not once per event** — In `_apply_6w_9m_shifts`, `_sample_non_pregnant_hemoglobin` calls `get_draw(index)`. In vivarium's randomness system, this is deterministic for the same index + additional_key, so multiple pipeline queries during the same event produce consistent values. This is correct but worth a comment confirming intentionality.

3. **IFA adjustment is still applied at postpartum events** — The `_adjust_exposure_for_ifa` modifier is registered without event gating. At postpartum events the hemoglobin pipeline still subtracts the IFA effect. For the 6w-9m event where hemoglobin is drawn from the non-pregnant distribution, applying a pregnancy-era IFA correction may not be appropriate. This may be pre-existing behavior, but the new postpartum logic makes it more visible.

4. **`_apply_6w_9m_shifts` applies shifts to indices not in `hgb`** — As noted in Design #2, `pph_mask` includes indices outside `survived_idx`. When `hgb.loc[pph_mask]` is evaluated, pandas aligns and ignores non-matching indices, so no incorrect data is written. However, if `pph_mask` contains `True` values for non-survived simulants, the `.loc[pph_mask]` selection on `hgb` silently returns an empty result for those indices rather than raising. This is functionally safe but masks potential logic errors.

### Minor Nits

1. hemoglobin.py: Section header comment uses `######` without a trailing blank line before the method definition — inconsistent with the earlier `########################` style that does have one.

2. data_values.py: The `(pregnant)` clarification on `ANEMIA_THRESHOLDS` is helpful but may now be incomplete given postpartum hemoglobin is in play. Consider `(pregnant and 0-6w postpartum)` if that's the usage scope.

### Overall

The core logic is sound and well-structured. The most important items to address are:

1. **`clip_quantiles` mutation** (Design #1) — easy fix, avoids a class of subtle bugs
2. **Index alignment in `_apply_6w_9m_shifts`** (Design #2) — compute masks on the survived subset for robustness
3. **DRY extraction** (DRY [#1-2](https://github.com/ihmeuw/vivarium_gates_mncnh/pull/319#1-2)) — low-risk refactor that consolidates 4 occurrences of the same pattern
4. **IFA adjustment at postpartum** (Functionality #3) — verify this is intentional before merging to main

Given this is marked as a sensitivity analysis/experiment (DO NOT MERGE), the lack of tests and changelog is acceptable for now, but should be addressed before promoting to a production branch.

src/vivarium_gates_mncnh/components/hemoglobin.py
```diff
@@ -65,6 +72,71 @@ def setup(self, builder: Builder) -> None:
...
+            modifier=self._modify_exposure_for_postpartum,
+        )
+
+    def _build_non_pregnant_distribution(self, builder: Builder) -> None:
+        """Build ensemble distribution machinery for non-pregnant hemoglobin sampling."""
+        non_pregnant_mean = builder.data.load(data_keys.HEMOGLOBIN.NON_PREGNANT_EXPOSURE)
+        sd_data = builder.data.load(data_keys.HEMOGLOBIN.STANDARD_DEVIATION)
+        raw_weights = builder.data.load(data_keys.HEMOGLOBIN.DISTRIBUTION_WEIGHTS)
+
+        # glnorm not currently supported by risk_distributions
```
patricktnast, line 103: feature request maybe??

src/vivarium_gates_mncnh/components/hemoglobin.py
```diff
@@ -65,6 +72,71 @@ def setup(self, builder: Builder) -> None:
...
+            )
+            for param_name, param_data in parameters.items()
+        }
+
+        self._non_pregnant_propensity_stream = builder.randomness.get_stream(
+            "non_pregnant_hemoglobin_propensity"
+        )
+        self._non_pregnant_ensemble_stream = builder.randomness.get_stream(
+            "non_pregnant_hemoglobin_ensemble_propensity"
+        )
```
zmbc: This should be using the same propensities as during pregnancy, is this doing that?

src/vivarium_gates_mncnh/components/hemoglobin.py
```diff
@@ -102,6 +185,114 @@ def on_time_step_cleanup(self, event: Event) -> None:
...
+        result = exposure.copy()
+        result.loc[survived_pop.index] = self._apply_hemorrhage_shifts(
+            result.loc[survived_pop.index].copy(),
+            survived_pop,
+            self.pph_shift_0_6w,
+            self.aph_shift_0_6w,
+        )
+        return result
+
+    def _apply_6w_9m_shifts(self, index: pd.Index, exposure: pd.Series) -> pd.Series:
```
patricktnast, line 262: it bugs the nitpick center of my brain that this and above are so similar, but i think refactoring loses as much as it gains. can you at least reorganize one to match the other more closely, though?

patricktnast (approved):

src/vivarium_gates_mncnh/components/hemoglobin.py
hussain-jafari: No! Will update

src/vivarium_gates_mncnh/components/hemoglobin.py
hussain-jafari: Yeah I tried to DRY this up but like you said there are enough small differences that it doesn't seem to be worth the tradeoff.

albrja (approved):

PR #324, opened 2026-06-02
add pop size to branches

hussain-jafari:
add pop size to branches


albrja (approved):
Why are you adding it here if you its only one value and not a list of values?

hussain-jafari:
From the ticket that Zeb made [MIC-7138]:

"We want to use 200k population size whenever we are not parallelizing the sim, e.g. when running the sim in an InteractiveContext as part of the hemoglobin -> maternal disorders PAF calculation. In order to make sure this happens, we should move the smaller population size to be part of the branches.yaml, since we only want that when we are running multiple (10) seeds."

patricktnast (approved):

rmudambi (approved):

PR #329, opened 2026-06-26
pin vbu<=3.3.2

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: [MIC-7261](https://jira.ihme.washington.edu/browse/MIC-7261)
- *Research reference*: 

### Changes and notes
 

Some of the dependencies (e.g. vivarium and possibly others) are pinned
to pre-monorepo versions which in turn pin vbu to its pre-monorepo version.
Until we update all pins to pre-monorepo version, just pin vbu down.

Ironically, vbu shouldn't even be getting pinned in the first place since it's
not used directly by the actual code base - it's only used in the makefile.
An alternative - and arguably better - solution here is to remove the vbu pin
from all framework repos!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,16 @@
 [build-system]
 requires = ["packaging", "setuptools"]
 
+# Minimal [project] block so vivarium_build_utils' base.mk can parse the
+# distribution name from pyproject.toml instead of falling back to
+# ``$(notdir $(CURDIR))`` (which is unsafe under Jenkins's parallel
+# workspaces — directories like ``<repo>@2`` contain ``@``, which uv
+# rejects in ``--config-settings-package`` values). Everything except
+# ``name`` continues to come from setup.py via the dynamic declaration.
+[project]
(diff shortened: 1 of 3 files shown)


hussain-jafari (approved):

patricktnast (approved):

setup.py
```diff
@@ -78,7 +81,8 @@
     ]
 
     setup(
-        name=about["__title__"],
+        # name is declared statically in pyproject.toml's [project] block.
```
patricktnast, line 84: is this comment needed?

pyproject.toml
```diff
@@ -1,6 +1,28 @@
 [build-system]
 requires = ["packaging", "setuptools"]
 
+# Minimal [project] block so vivarium_build_utils' base.mk can parse the
```
patricktnast, line 4: on this point, i get avoiding scope creep, but is it very difficult to swap to pyproject.toml wholesale? if not here, do we have a ticket for that?

PR #334, opened 2026-07-06
Monorepo migration test updates

hussain-jafari:
### Description
- *Category*: test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7250

### Changes and notes
- Updated env to point at testing utils >= 0.7.4 so we use the enterprise vivarium inputs.
- Updated interactive hemoglobin test so that our assert statements compare DataFrames with consistent index orders.
- Reran all tests including notebooks.

### Verification and Testing
Tests pass.

### Next steps

- [X] Merge this pull request


setup.py
```diff
@@ -96,10 +98,8 @@
             "data": data_requirements
             + cluster_requirements
             + lint_requirements
-            + test_requirements,
-            # TODO: fold validation_requirements back into "data" once a
-            # vivarium_testing_utils release supports vivarium_inputs>=8.0.0
-            # (0.7.1's [validation] extra caps vivarium_inputs<8.0.0).
+            + test_requirements
+            + validation_requirements,
             "validation": validation_requirements,
```
patricktnast, line 103: do we still need the extra namespace if we put it back into "data"?

patricktnast:
we have an assert failure that didn't exist before, and a few new NaNs in some of the dataframes. is this expected?

hussain-jafari:
> we have an assert failure that didn't exist before, and a few new NaNs in some of the dataframes. is this expected?

This is in an interactive notebook which is used as a reference, and not in the executed notebook which contains the actual final outputs. What probably happened was I ran the interactive notebook in an intermediate buggy state when debugging and then forgot to clear the outputs. Clearing the outputs of the interactive notebook now!

setup.py
hussain-jafari: I suppose not. Let me confirm with Jim why this was originally split out.

albrja (approved):
Im approving based on the setup update which I don't will matter in a week or two. I can't speak as to the results notebook updates.

patricktnast (approved):

PR #335, opened 2026-07-08
Adjust APH/PPH temporal ordering and data updates (MIC-7286)

hussain-jafari:
### Description
- *Category*: implementation (+ data artifact, observers, test)
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7286
- *Research reference*: https://github.com/ihmeuw/vivarium_research/pull/1946 (Adjust temporal ordering of maternal disorders)

### Changes and notes
Implements the APH/PPH split temporal-ordering + data updates from research PR #1946.

**Artifact / data loading**
- APH incidence risk is now **per birth** — `(1 − postpartum_fraction)·incidence_c367 / birth_rate` — dropping the abortion/miscarriage (c995) + ectopic (c374) terms from the old pregnancy-rate denominator.
- PPH incidence risk divides by `birth_rate − antepartum_hemorrhage_csmr` (per surviving birth).
- New artifact key `cause.antepartum_hemorrhage.cause_specific_mortality_rate = (1 − postpartum_fraction)·csmr_c367`.

**Component**
- Maternal mortality is split into **two passes**: an antepartum pass (APH + abortion/miscarriage/ectopic) at a new `antepartum_maternal_disorders_mortality` step, then an intrapartum pass (sepsis / obstructed labor / PPH / residual) at `mortality` over `mother_alive` survivors only. APH death and intrapartum death are therefore mutually exclusive.
- APH restricted to full-term (still/live) pregnancies; intrapartum-disorder incidence gated on `mother_alive`; sepsis/OL incidence and residual CFR denominators become `birth_rate − aph_csmr`.
- Hardening: a `setup`-time guard that the event schedule assigns each disorder before its mortality pass, and a structural `MATERNAL_DISORDER_PHASE` mapping for the antepartum/intrapartum partition.

**Observer**
- Residual `ylds_per_case` denominator becomes `birth_rate − aph_csmr`; all other maternal-disorder outputs flow unchanged (they read persistent columns at the `mortality` step).

A shared `load_births_net_of_aph_mortality(builder)` helper backs the three component denominators.

### Verification and Testing
Verified against a rebuilt Ethiopia artifact (draw 60). **The subset copy+recompute artifact matched a full GBD rebuild to floating-point precision (max relΔ ~1e-14)**, and the refactored loaders match the stored artifact (max|Δ| ~1e-16).

InteractiveContext checks (`tests/test_hemorrhage_temporal_ordering.py`) — 100k, all pass:

| Check | Result |
|---|---|
| APH incidence by outcome | live 8.58%, still 8.98%, partial_term **0%** |
| APH per-birth vs per-pregnancy proxy | **1.6–2.1× higher** (denominator dropped AME terms) |
| Intrapartum risk bump `/birth_rate → /(birth_rate−aph_csmr)` | ratio ~1.001–1.002 |
| APH deaths vs intrapartum deaths | **disjoint**; 0 intrapartum flags among APH-dead |
| AME only on partial_term; all APH & PPH deaths severe | pass |

Trace plots/tables: `tests/model_notebooks/interactive/executed/interactive_simulation_aph_temporal_ordering.ipynb`.

**Accepted behavior (confirmed with research):** mothers who die of antepartum hemorrhage still give birth — i.e. dead-of-APH dyads intentionally continue through the delivery/intervention/neonatal components. This is by design, not a defect.

**Deferred to a follow-up:** added unit-test coverage for the new data formulas (`load_antepartum_hemorrhage_csmr`, the per-surviving-birth denominators, `calculate_case_fatality_rates`, the observer residual `ylds_per_case`).

- [x] model results directory is up to date (`MODEL_RESULTS_DIR = model39.0`)
- [ ] all tests pass (`pytest --runslow` with both *vivarium_gates_mncnh_artifact* and *vivarium_gates_mncnh_simulation*) — **requires the artifact placed at the model39.0 path above**

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/vivarium_gates_mncnh/components/maternal_disorders.py
```diff
@@ -93,15 +101,12 @@ def on_time_step(self, event: Event) -> None:
...
-            .set_index("year_start")
-            .drop(columns=["year_end"])
-            .reindex(asfr.index, level="year_start")
-        )
-        birth_rate = (sbr + 1) * asfr
-        incidence_risk = (raw_incidence / birth_rate).fillna(0.0)
+        # This shared loader divides by births net of antepartum-hemorrhage deaths,
+        # correct for the intrapartum subclasses (sepsis, obstructed labor) that
+        # consume this pipeline. AbortionMiscarriageEctopicPregnancy also inherits it
+        # but assigns deterministically off pregnancy outcome and never reads the pipeline.
```
patricktnast, line 107: if this is the case, can we do the same override there that ResidualMaternalDisorders does?

zmbc:
> Accepted behavior (confirmed with research): mothers who die of antepartum hemorrhage still give birth — i.e. dead-of-APH dyads intentionally continue through the delivery/intervention/neonatal components. This is by design, not a defect.

I thought the opposite, though I may be misremembering; do you have a pointer to where this was confirmed?

hussain-jafari:
> 

I talked to Ali and Tyler before running the model and confirmed that this is what we want, and that it will be documented but is not currently. 

> > Accepted behavior (confirmed with research): mothers who die of antepartum hemorrhage still give birth — i.e. dead-of-APH dyads intentionally continue through the delivery/intervention/neonatal components. This is by design, not a defect.
> 
> I thought the opposite, though I may be misremembering; do you have a pointer to where this was confirmed?

I talked to Ali and Tyler before running the model and confirmed that this is what we want, and that it will be documented but is not currently.

tests/model_notebooks/interactive/executed/interactive_simulation_hemorrhage.ipynb
```diff
@@ -3,16 +3,7 @@
   {
```
tylerdy, line 3: <p><span class="mention" data-value="hussain-jafari" data-denotation-char="@">﻿<span contenteditable="false">@hussain-jafari</span>﻿</span> From these comments and the table it seems that with APH incidence risk being per birth, there is now more of a difference between the APH incidence for misprostol_available vs unavailable groups. Misoprostol availability is those who attend ANC and have a home birth right? For misoprostol availability tracking ANC/hemoglobin I assume you're referring to <a href="https://vivarium-research.readthedocs.io/en/latest/models/risk_effects/hemoglobin/index.html#maternal-disorders" target="_blank">this</a> about custom PAFs for hemoglobin on maternal disorders:</p><blockquote>The PAF calculation does not take detailed correlation with other factors that affect postpartum hemorrhage and maternal sepsis risk into account. For instance, the correlation between ANC attendance and in-facility delivery in our simulation will induce a correlation between hemoglobin (as a result of baseline IFA coverage distributed at ANC) and misoprostol coverage (available at home deliveries only), both of which affect postpartum hemorrhage risk. We hypothesize that the impact of ignoring this sort of correlation will be small enough for our model to calibrate, but we can revisit this assumption if the model calibrates poorly.</blockquote><p>Do you have a theory on why excluding a/e/m pregnancies would widen this APH incidence difference bt misoprostol available and unavailable? cc <span class="mention" data-value="zmbc" data-denotation-char="@">﻿<span contenteditable="false">@zmbc</span>﻿</span> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/335/discussion/#comment-3751492919'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

tylerdy: <p>Like is it just because there is a smaller denominator now and the correlation only exists for FT births or something? Wanted to ask about this bc I hadn't thought of this hemoglobin/miso avail tracking before when I wrote this test so having trouble thinking through it. </p><p> </p>

zmbc: <p>I have a feeling that this may be a result of the misoprostol_available column defaulting to False for abortion/miscarriage/ectopic pregnancies and those not being filtered out of this table. They should be filtered out, and then we should put the check back to what it was, I think.</p>

zmbc: <p>(Previously we included them because they were at risk of APH, but that is no longer the case.)</p>

zmbc: And if we don't already have a check that no abortion/miscarriage/ectopic pregnancies have incident APH, we should add that!

tylerdy: <p>a/m/e pregnancies should already be filtered out of the aph_incidence values here, since they are no longer included in aph_count, and the denominator is already the count of full term pregnancies. To double check I adjusted FT pregnancies to also exclude "INVALID_OUTCOME" pregnancy outcomes but the results didn't change (there's only 'partial_term', 'live_birth', 'stillbirth' in the pop)</p><p> </p><p>I added an assert here for no a/m/e APH, which passes</p>

tylerdy: <p>Resolved in MNCNH check-in- the commented explanation for the difference in misoprostol available vs unavailable makes sense and the new check is good. We decided to remove APH component based on the placentia previa discussion we just had, but if we were keeping APH I would update the old check to compare only within at-home birth populations. </p>

patricktnast (approved):

PR #351, opened 2026-07-23
Temp/aph pph shifts

hussain-jafari:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-XYZ
- *Research reference*: 

### Verification and Testing

<--
*** REMINDER ***
CI WILL NOT RUN ANY TESTS.
-->
- [ ] all tests pass (`pytest --runslow` with both *vivarium_gates_mncnh_artifact* and *vivarium_gates_mncnh_simulation*)

### Next steps

- [ ] Merge this pull request
- [ ] THIS IS A SENSITIVITY ANALYSIS/EXPERIMENT -- DO NOT MERGE!


src/vivarium_gates_mncnh/model_specifications/model_spec.yaml
```diff
@@ -107,17 +110,19 @@ configuration:
...
             - 'mortality'
+            - 'early_postpartum'
+            - 'late_postpartum'
             - 'early_neonatal_mortality'
             - 'late_neonatal_mortality'
             - 'postpartum_depression'
 
     population:
-        population_size: 200_000
+        population_size: 20_000
```
tylerdy: @hussain-jafari Is there a reason the population size compared to main keeps being different when we do these model runs? This makes it a lot harder to do V&V. Would be helpful if we could figure out a way to always have main be the same population size and that is the size thats used in model runs unless requested otherwise.

tylerdy: The population size also didnt look like it changed for the branch with main merged in https://github.com/ihmeuw/vivarium_gates_mncnh/pull/376/changes#diff-ee69bfe0cbc968156b60357752c0422e5671d5816f002ec7b7b3a62be6b5b213 so I'm a bit confused why that one also has a different population size than main

tylerdy:
This and the branch with the changes from main (https://github.com/ihmeuw/vivarium_gates_mncnh/pull/376) look good to me. Before we merge these into main we need to make sure they are the standard population size so that we have a consistent population size to compare between main and future model runs that use the standard population size. See above comment

PR #353, opened 2026-07-24
Separate LBWSG-affected causes in neonatal mortality risk

hussain-jafari:
### Description
- *Category*: implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7320
- *Research reference*: https://github.com/ihmeuw/vivarium_research/pull/1760 — [neonatal cause model, ACMRisk LBWSG decomposition](https://vivarium-research.readthedocs.io/en/latest/models/causes/neonatal/index.html#cause-neonatal-disorders-mncnh)

### Changes and notes

Separates the LBWSG **causal** effect from its **non-causal** (confounded) association in the
neonatal all-cause mortality-risk (ACMRisk) calculation, per research PR #1760. Previously the
scenario LBWSG relative risk scaled *all* of ACMRisk, so a birthweight/gestational-age
intervention wrongly moved mortality from causes LBWSG doesn't cause (e.g. congenital anomalies).

Now `ACMRisk = affected·(1−PAF)·RR(scenario exposure) + unaffected·(1−PAF)·RR(baseline exposure)`,
where `affected = Σ CSMRisk` over the 12 GBD-2021 LBWSG-affected causes and
`unaffected = ACMRisk − affected`. Interventions therefore move only the affected fraction.

Layered commits:
1. **Artifact loader** — new key `cause.all_causes.affected_causes_mortality_risk` (sum of neonatal
   deaths over the 12 affected GBD causes / adjusted birth counts).
2. **Component** — the decomposition in `NeonatalMortality.get_acmr_pipeline`; a new pre-intervention
   ("unmodified") LBWSG exposure and a baseline all-cause RR pipeline in `LBWSGRiskEffect`; the
   all-cause target skips its uniform RR modifier so the RR is applied non-uniformly. The
   modeled-subcause subtract/add-back is unchanged.
3. **Review fixes** — clamp `unaffected ≥ 0`; batch the loader's GBD pull (24 calls → 2); DRY the RR
   helpers; docstrings.
4. **Tests** — white-box unit tests for the decomposition arithmetic + a gated InteractiveContext
   check file.

Guidance for reviewers: the interesting file is `components/lbwsg.py` (the baseline-RR pathway and
the `register_target_modifier` skip). No `vivarium`/`vivarium_public_health` changes. Implementation
was checked line-by-line against the merged research docs.

**Note on `other_causes` vs. the algebraic `unaffected` term:** the sim only models 4 of the 12
affected causes; the other 8 (diarrhea, LRI, meningitis, SIDS, …) stay lumped in `other_causes` at
the *scenario* RR, so observable `other_causes` is not identical to the frozen `ACMRisk_unaffected`
and moves slightly under intervention. This matches the spec; see the verification below.

### Verification and Testing

Artifact rebuild is **surgical** for review: the new key was appended to a private copy of
`model33.1/ethiopia.hdf` (existing keys skipped). Against **real GBD**, the loader batched correctly
and `affected ≤ all_cause` held in **0/1000** draw-cells.

- **Unit tests** (`tests/test_neonatal_mortality_decomposition.py`, no artifact needed): **16 pass**
  — exact decomposition arithmetic, the clamp, and the identity-collapse (`scenario RR == baseline RR`
  → old `acmr·(1−PAF)·RR`).
- **Behavioral sim-verify** (baseline vs `mms_total_scaleup`, 100k, common random numbers) — all four
  research expectations hold:

  | | affected rate | other-causes rate |
  |---|---|---|
  | baseline | 0.01758 | 0.01118 |
  | mms_total_scaleup | 0.01618 | 0.01098 |

  - Affected mortality **responds to MMS** (−7.9%); other-causes **~invariant** (−1.8%, the expected
    small residual from the 8 non-modeled affected causes).
  - LBWSG gradient (baseline): affected rate `BW<2500g` 0.0818 vs `BW≥2500g` 0.0061; other-causes
    0.0439 vs 0.0053 — both vary with exposure, only affected responds to intervention.

**Known limitation (follow-up, not a blocker):** the gated checks in
`tests/test_lbwsg_affected_causes.py` build on the pre-existing `tests/utilities.py`
`get_births_and_deaths_idx` helper (also behind the already-`@skip`ped `test_neonatal_acmr` /
`test_neonatal_csmr`), whose neonatal age-window doesn't match the interactive states, so they don't
yet pass against a keyed artifact. Repairing that shared harness + un-skipping the neonatal V&V is a
separate follow-up.

- [ ] all tests pass (`pytest --runslow` with both *vivarium_gates_mncnh_artifact* and *vivarium_gates_mncnh_simulation*)

### Next steps

- [ ] Rebuild the production artifact with the new key (`make_artifacts`, all locations) before a run request
- [ ] Follow-up: fix `get_births_and_deaths_idx` and un-skip neonatal V&V so the gated checks run
- [ ] Merge this pull request

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/vivarium_gates_mncnh/components/lbwsg.py
+++ b/src/vivarium_gates_mncnh/components/lbwsg.py
@@ -205,6 +205,11 @@ class LBWSGRisk(LBWSGRisk_):
 
     AXES = [BIRTH_WEIGHT, GESTATIONAL_AGE]
 
+    @staticmethod
+    def get_unmodified_exposure_name(axis: str) -> str:
+        """Return the pre-intervention (unmodified) exposure column name for an axis."""
+        return f"{axis}.unmodified_exposure"
+
     def __init__(self):
         super().__init__()
         self.continuous_propensity_column_name = {
@@ -226,6 +231,13 @@ class LBWSGRisk(LBWSGRisk_):
             initializer=self._initialize_continuous_propensities,
             columns=[self.continuous_propensity_column_name[axis] for axis in self.AXES],
             required_resources=[self.randomness],
+        )
+
+        # Pre-intervention (unmodified) exposure columns. Assigned at ultrasound alongside
+        # the modified ``*.exposure`` columns; initialized to NaN here.
+        builder.population.register_initializer(
+            initializer=self.initialize_unmodified_exposure,
+            columns=[self.get_unmodified_exposure_name(axis) for axis in self.AXES],
         )
 
     #################
@@ -255,6 +267,18 @@ class LBWSGRisk(LBWSGRisk_):
         exposures = pd.DataFrame(
             {
                 self.get_exposure_name(axis): pd.Series(np.nan, index=pop_data.index)
+                for axis in self.AXES
+            }
+        )
+        self.population_view.initialize(exposures)
+
+    def initialize_unmodified_exposure(self, pop_data: SimulantData) -> None:
+        """Initialize unmodified exposure columns to NaN; assigned at ultrasound."""
+        exposures = pd.DataFrame(
+            {
+                self.get_unmodified_exposure_name(axis): pd.Series(
+                    np.nan, index=pop_data.index
+                )
                 for axis in self.AXES
             }
         )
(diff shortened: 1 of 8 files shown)


patricktnast (changes requested):
9c5beb5 wasn't quite what i was expecting--my hope was that we could get rid of `get_baseline_relative_risk_column_name` and column_name_fn, which is ultimately what i want by splitting apart the LBWSGRiskEffects for scenario and baseline in such a way that for the baseline risk effects, we just *overwrote* `get_relative_risk_column_name(age_group)` to include the prefix, so we didn't have to replace it with `column_name_fn` in lbwsg.py in `_get_relative_risk`. then you don't need the baseline functions, _recompute_relative_risk, _get_age_resolved_rr. 

I think that would be marginally smaller lbwsg code than your previous single-component version; otoh, the previous was OK too. but this is in a non-ideal middle ground between the two.

src/vivarium_gates_mncnh/components/lbwsg.py
```diff
@@ -481,58 +544,137 @@ def get_relative_risk_for_age_group(age_group: str) -> pd.Series:
...
+        return LBWSGRisk.get_unmodified_exposure_name(BIRTH_WEIGHT)
+
+    @property
+    def gestational_age_exposure_column_name(self) -> str:
+        return LBWSGRisk.get_unmodified_exposure_name(GESTATIONAL_AGE)
+
+    def get_relative_risk_column_name(self, age_group: str) -> str:
+        return f"baseline_{super().get_relative_risk_column_name(age_group)}"
+
+    def setup(self, builder: Builder) -> None:
```
patricktnast: I think you could incorporate this into the parent (LBSWGRISK) class's setup by doing the following:
```
def setup(self, builder: Builder) -> None:
    self._sim_step_name = builder.time.simulation_event_name()
    self.paf_pipeline_name = f"lbwsg_paf_on_{self.target.name}.{self.target.measure}.paf"
    # RiskEffect.setup rather than vph's LBWSGRiskEffect.setup: vph declares the RR
    # initializer against the exposure *pipeline* and "sex" (the mother's column), but
    # initialize_relative_risk reads the state-table columns named by
    # lbwsg_exposure_column_names, plus sex_of_child. Declaring what we read is what
    # lets a subclass retarget the whole track by overriding the exposure names.
    self.age_intervals = self.get_age_intervals(builder)
    RiskEffect.setup(self, builder)
    self.interpolator = self.get_interpolator(builder)
    builder.population.register_initializer(
        initializer=self.initialize_relative_risk,
        columns=self.rr_column_names,
        required_resources=self.lbwsg_exposure_column_names + [COLUMNS.SEX_OF_CHILD],
    )
    self.register_paf_pipeline(builder)

  def register_paf_pipeline(self, builder: Builder) -> None:
      """Expose this target's LBWSG PAF as a pipeline for other components to read."""
      # VPH 5.1 no longer builds ``self.paf_table`` on RiskEffect; build it here
      # from the PAF data the base now loads onto ``self.paf_data``.
      self.paf_table = self.build_lookup_table(
          builder, "population_attributable_fraction", data_source=self.paf_data
      )
      builder.value.register_attribute_producer(
          self.paf_pipeline_name,
          source=self.paf_table,
          required_resources=[self.paf_table],
      )
```
and then have the baseline risk producer just no-op the register paf pipeline. not a big deal though

patricktnast (approved):

PR #358, opened 2026-08-03
Run V&V notebooks for model 39.0 and fix asserts

hussain-jafari:
### Description
- *Category*: test
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7346
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/intervention_models/mncnh_pregnancy/iv_iron_antenatal/iv_iron_mncnh.html

### Changes and notes

Runs the full V&V notebook suite against the **model 39.0** results and fixes the
two asserts that failed. Stacked on `hjafari/feature/mic-7286_aph_pph_temporal_ordering`
(#335), so the diff here is only the notebooks.

**Two assert fixes in `tests/model_notebooks/results/hemoglobin.ipynb`** — the only
source change in this PR:

1. *Cross-scenario pregnancy-outcome check.* `full_term` is a transient
   `pregnancy_outcome` that `pregnancy.py` resolves into `live_birth`/`stillbirth`
   before observation, so it is an always-empty categorical level: its baseline
   proportion is 0 and the cross-scenario ratio is `0/0 = NaN`. Added `.dropna()`
   so the check compares only realized outcomes.
2. *IV-iron neonatal-death check.* IV iron raises hemoglobin, which lowers neonatal
   sepsis mortality (MIC-7289), so the IV iron scenario now has strictly fewer
   neonatal deaths than baseline. The assert demanded *equality*, on the premise
   that IV iron had no neonatal pathway; relaxed to a strict decrease and updated
   the stale comment.

Equivalent fixes landed on the 37.1 lineage in `c60c8ea` but were never ported to
the model-39 line — `git diff v39.0 <branch> -- tests/model_notebooks/results/hemoglobin.ipynb`
is empty for both branches containing the tag. Cherry-picking `c60c8ea` wholesale
would drag in its post-monorepo imports and combined `maternal_hemorrhage_counts`
read, neither of which applies to the split-era model-39 code, so only the two
relevant hunks were ported.

**Refreshed executed mirrors** (10 notebooks) from this run. Reviewers can skip
the output churn — all ten re-run with 0 error outputs and no unrun cells.

### Verification and Testing

Full suite run on the model 39.0 results, in both environments:

| Set | Env | Collected | Outcome |
|---|---|---|---|
| Non-results | simulation | 27 | 19 passed, 8 skipped |
| Non-results | artifact | 13 | 13 passed |
| Results | artifact | 5 | 5 passed |

Totals reconcile: 45 collected, **37 passed, 8 skipped, 0 failed, 0 errored**.

Two things worth flagging rather than burying:

- **The 8 skips are pre-existing, not new.** `test_neonatal_acmr` and
  `test_neonatal_csmr` carry `@pytest.mark.skip(reason="This test is not currently
  working due to issues with the mortality compoents.")` in v39.0's own source. The
  neonatal ACMR/CSMR checks were therefore never actually exercised.
- **Results are 5/5 only with the two fixes in this PR.** v39.0 exactly as tagged is
  4/5 — `hemoglobin` fails on the `full_term` NaN assert and then on the IV-iron
  equality assert.

Separately: the results set only ran at all after the double-nested
`model39.0/<loc>/<loc>/` results directories were flattened on disk. That was a
data-layout problem, not a code one, and needs no change in this repo.

- [x] model results directory is up to date (`MODEL_RESULTS_DIR = "model39.0"`)

### Next steps

- [x] Merge this pull request


tests/model_notebooks/interactive/executed/interactive_simulation_gestational_age.ipynb
```diff
@@ -3,19 +3,19 @@
   {
```
patricktnast, line 3: <p>huh, why did these versions go *down*?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/358/discussion/#comment-3716225434'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

tests/model_notebooks/results/hemoglobin.ipynb
```diff
@@ -2718,8 +2718,12 @@
    "metadata": {},
```
patricktnast, line 2718: <pre class="ql-syntax" spellcheck="false">Line #6.        [x for x in scenarios_run if x not in ['mms_total_scaleup', 'anemia_screening_and_iv_iron_scaleup']])].dropna() == 1).all(), "non-MMS/IV iron scenario changing pregnancy outcomes"
</pre><p>the reasoning seems plausible, but I don't really understand why this would have changed one way or the other?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/358/discussion/#comment-3716225438'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='161'/>

patricktnast (approved):
I'm just assuming the non-outputted plots are OK.

tests/model_notebooks/results/hemoglobin.ipynb
hussain-jafari: I believe this was actually a change from pandas.

zmbc: <p>I feel like I have seen this code added in multiple PRs, right? So it's probably best to keep it exactly the same in all of them and reduce our merge headaches. But, as a nit, it would be nicer to drop full_term explicitly, which would obviate the comment and be less likely to hide bugs than <code>dropna</code></p>

zmbc (changes requested):
The fact that the plots are not displaying seems like a problem

zmbc:
Hey @hussain-jafari I think the objective of this PR may have already been done by Tyler in #357?

hussain-jafari:
plots are restored! @zmbc 

> Hey @hussain-jafari I think the objective of this PR may have already been done by Tyler in #357?

@zmbc oh true! I'll close this then.

PR #359, opened 2026-08-04
Don't apply iron effects to non-pregnant hemoglobin

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7360
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/postpartum_hemoglobin_module/module_document.html

### Changes and notes

Updates the non-pregnant hemoglobin exposure so the iron interventions are no longer
applied to it. Two commits, reviewable independently.

**`62baa50` — define the non-pregnant exposure.** At `late_postpartum` the hemorrhage
shifts overwrite `hemoglobin.exposure` with a shifted and clipped version of a fresh
draw from the non-pregnant distribution. That draw was a local inside
`_apply_6w_9m_shifts`, so it had no identity of its own. Registers it as the attribute
`non_pregnant_hemoglobin.exposure`, and has `_apply_6w_9m_shifts` consume the attribute
rather than call the sampler. The sampler is a pure function of the stored propensities,
so it is deterministic across calls and can be evaluated for the whole population, not
just the survivors the simulation assigns it to.

**`ac50ca6` — gate the iron modifiers off after six weeks.**
`apply_oral_iron_to_hemoglobin` and `apply_iv_iron_to_hemoglobin` had no event gating,
so they ran at *every* event including `late_postpartum`, re-adding their effects on top
of the fresh non-pregnant draw — while the wholesale replacement simultaneously
discarded `_adjust_exposure_for_ifa`'s risk-deletion term. Post-six-week hemoglobin was
therefore the inverse of spec on both counts.

The module document is explicit:

> We assume that the impacts of all interventions that have modified hemoglobin during
> pregnancy last only until six weeks after the end of pregnancy.

It never mentions IFA or calibration for the post-six-week period; the deletion formula
(`gbd_hemoglobin_exposure - baseline_ifa_overall * ifa_hemoglobin_shift`) is scoped to
the *start of pregnancy*. 0-6w is untouched by this change — interventions remain in
force there, arriving by inheritance from hemoglobin at birth, as the spec intends.

### Verification and Testing

Interactive run, draw 60, 20k pop, `artifacts/aph/ethiopia.hdf`, 11276 survivors:

| check | result |
|---|---|
| 6w-9m reconstructs as `clip(non_pregnant + shifts, 0)` | max err **1.4e-14** |
| …with iron terms added back (must now fail) | off by **9.9223** |
| per-simulant shift recovery at 6w-9m | **1555/1555 exact**, 0 false positives |
| non-pregnant exposure deterministic on resample | **True** |
| 0-6w iron still applied (regression) | **5191** treated |
| 0-6w shift via `mortality`→`early_postpartum` differencing | **1551/1551 exact** (4 clipped excluded) |

Scale of the bug being fixed: **5191 of 11276 survivors** were affected, leaking
9.92 g/L — larger than the 6w-9m APH shift itself (−5.88). Pre-existing since v38.1
(verified by reproducing it on pristine `eb5d470`), so **model 38.x results carry it**.

### Next steps

- [X] Merge this pull request


patricktnast (approved):

src/vivarium_gates_mncnh/components/intervention.py
```diff
@@ -920,6 +929,13 @@ def setup(self, builder: Builder) -> None:
     def apply_iv_iron_to_hemoglobin(
         self, index: pd.Index, exposure: pd.Series[float]
     ) -> pd.Series[float]:
+        # Pregnancy interventions stop mattering six weeks after the end of
```
zmbc, line 932: Yeah the code is aligning with my concern; I'm not seeing where we are avoiding applying the baseline IFA deletion to non-pregnant hemoglobin.

hussain-jafari: Will update and launch runs.

PR #360, opened 2026-08-04
Correct IFA birth-weight deletion shift

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7350

### Changes and notes
`risk_specific_shift` is the baseline deletion shift: it removes the average IFA birth-weight effect already embedded in the GBD exposure data so that adding each simulant's own effect doesn't double count. Per the oral iron model docs it is `baseline_ifa_overall_draw * ifa_shift_draw` -- a function of the *same* effect-size draw -- and the resulting calibration should "preserve the population mean value of birthweight".

The birth-weight term passed `key` to `load_excess_shift`, and `key` here is the risk_specific_shift key. `get_random_variable` seeds on the artifact key name, so that returned an independent draw of the same distribution (norm, mean 57.73, sd 25.5) rather than the effect actually applied. At draw 60 the effect is 9.26 g while 59.43 g was being removed, so nothing cancelled and mean birth weight was not preserved.

Request EXCESS_SHIFT explicitly so the deletion shift is derived from the same draw, and drop the RISK_SPECIFIC_SHIFT entry from load_excess_shift's distribution dict so the same mistake raises `Unrecognized key` instead of silently returning a different number.

Regressed in 692151d (PR #244), which split one coverage-weighted calculation into separate GA and BW terms; the GA half kept its weighting, the BW half was rewritten without it and under the wrong key. MIC-7153 (9b948b2) restored the weighting but not the shared draw, so this completes that fix.

### Verification and Testing
Verified against Ethiopia: load_risk_specific_shift now returns birth_weight 4.273491, matching the pre-regression artifacts (model24.0 2025-12-08 and model28.0 2026-02-04), and gestational_age is unchanged at 0.064106. load_excess_shift rejects the risk_specific_shift key and still resolves both EXCESS_SHIFT keys.

<--
*** REMINDER ***
CI WILL NOT RUN ANY TESTS.
-->

### Next steps

- [X] Merge this pull request


zmbc (approved):
LGTM

src/vivarium_gates_mncnh/data/loader.py
```diff
@@ -1447,7 +1447,11 @@ def load_risk_specific_shift(
         birth_weight_shift = (
             (
                 exposure
-                * load_excess_shift(key, location)[excess_shift.columns]
+                # Request the EXCESS_SHIFT key, not `key`: get_random_variable seeds on the
```
zmbc: not sure how important this comment is

patricktnast (approved):

tests/model_notebooks/results/executed/neonatal.ipynb
zmbc, line 1: Most of the diff here appears to be because 33.1 tests were not re-run with the latest version of this notebook

zmbc (approved):
Overall nothing concerning in V&V, I noted a few small issues in the notebooks but those don't need to be addressed here

tylerdy:
<p>actually probably 36.2 not 33.1, the results tests here just have model_dir=33.1 so it was confusing</p>

---

 <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/360/discussion/#comment-5283484074'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #368, opened 2026-08-12
Merge epic/sepsis_on_hemoglobin into main

hussain-jafari:
### Description
- *Category*: implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7154
  (also carries MIC-7297 postpartum periods and MIC-7347 anemia-YLD timestep stratification)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021_mncnh/index.html

### Changes and notes

Merges `epic/sepsis_on_hemoglobin` into main and ports it onto the monorepo stack
(vivarium-engine 5.3.x / VPH 6.x). The epic branched before the monorepo
migration, so most of the work here is conflict resolution and porting rather
than new modelling.

**The model change.** `SepsisEffectsOnHemoglobin` applies a hemoglobin shift to
simulants who had maternal sepsis, with separate early-postpartum (0-6 wk) and
late-postpartum (6-39 wk) magnitudes. This adds two `simulation_events`,
`early_postpartum` and `late_postpartum`, and moves anemia-YLD accrual onto them
(it previously used a single six-week window measured at
`early_neonatal_mortality`).

**Conflict resolutions worth flagging:**

- **Data CSVs (100 files).** Took the epic's values. Both sides regenerated, but
  main's differ from the merge base only at ~1e-6 (re-run noise) while the epic's
  differ at ~1e-2. These data are functions of the maternal hemoglobin
  distribution, which this epic changes, so the epic's regeneration is the one
  the model change requires.
- **Deleted `tests/test_sepsis_hemoglobin.py`.** It reached main only as an import
  port during the neonatal merge; the epic superseded it and then deleted the
  smoke tests outright in `14c8daf`. It also asserted nothing useful — its only
  check is that the sepsis group's mean hemoglobin is lower, which is already
  true without the component, since `HemoglobinRiskEffect` makes low hemoglobin
  *cause* sepsis. Measured: -12.57 g/L on main with no component wired vs
  -15.47 g/L here, so it passed either way. **See "follow-up" below.**
- **`results/hemoglobin.ipynb` IV-iron check kept.** nbdime flagged it
  `REMOTE CELL DELETED`, but the cell is absent from the merge base and from every
  epic commit — main added it in `fd20996` after the epic branched, so the marker
  was a cell-alignment artifact, not an epic decision.
- **Two latent stack bugs fixed.** `interactive_simulation_sepsis_on_hemoglobin.ipynb`
  was authored on the old stack and never ported, so it failed on its first cell.
  `interactive_simulation_hemoglobin.ipynb` needed `.astype(float)` restored —
  `anemia_status` comes from `pd.cut`, and pandas 2.x preserves Categorical
  through `Series.map`, so `Categorical * float` raises. Both passed on the epic's
  older stack and only broke here.

### Verification and Testing

**Artifact.** `model42.1` = `model42.0` plus
`cause.maternal_sepsis_and_other_maternal_infections.hemoglobin_shift`, which
model42.0 lacked (without it every simulation test errors at component setup).
Verified per location as 116 keys vs 115 — one added, none removed or changed.
The shift is location-independent by construction, so all three locations carry
identical values; draw 60 gives -13.2479 early / -3.0423 late.

**V&V run — `model42.1`, 2,100 / 2,100 jobs succeeded.**

| location | jobs | duration |
|---|---|---|
| ethiopia | 700/700 | 9m28s |
| nigeria | 700/700 | 9m41s |
| pakistan | 700/700 | 6m14s |

700 jobs = 10 draws x 10 seeds x 7 scenarios; at `population_size: 20_000` per
branch that is 200k simulants per draw per scenario. Each run's recorded
`model_specification.yaml` confirms the `model42.1` artifact and
`SepsisEffectsOnHemoglobin`, i.e. workers ran the merged code.

**Component is doing what it should.** Sepsis/no-sepsis hemoglobin gap measured
against the artifact's own targets:

| | early step | neonatal steps |
|---|---|---|
| this branch | -18.51 | -15.47 |
| main, component absent | — | -12.57 |
| artifact shift, draw 60 | -13.25 | -3.04 |

The residual gap on main is the reverse-causation confound described above.

**Test suites.** Both were run, since `conftest.py` makes them mutually exclusive
by environment (the simulation env ignores `model_notebooks`, the artifact env
ignores `tests/test_*.py`) — a single invocation cannot cover both.

- artifact env: **22 passed, 1 skipped, 0 failed** (all 14 notebooks, in one pass)
- simulation env: **7 passed, 8 skipped, 0 failed** (on final HEAD)

The 8 skips in the simulation env are pre-existing mortality-component skips,
unchanged from main; the 1 skip in the artifact env is the empty `artifact/`
notebook directory.

`executed/` notebooks were regenerated in a single pass with both envs pointed at
this branch and `MODEL_RESULTS_DIR=model42.1` — the state they will run in once
this merges. Results notebooks each lose two cells: the papermill `parameters`
cell this branch removes, plus the injected-parameters cell papermill no longer
needs now that `model_dir` is read from `constants/paths.py`.

- [x] model results directory is up to date (`model42.1`)

### Next steps

- [X] Merge this pull request


zmbc:
We probably didn't want to merge this to `main` without checking V&V

tests/model_notebooks/results/executed/facility_choice.ipynb
```diff
@@ -5,10 +5,10 @@
    "id": "42c4fb6b",
```
zmbc, line 5: <p>Thanks!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/368/discussion/#comment-3770196318'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

PR #369, opened 2026-08-12
Redraw partial-term pregnancies at late postpartum

hussain-jafari:
### Description
- *Category*: bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7360
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/postpartum_hemoglobin_module/module_document.html

Stacked on #359 — review that one first. This PR depends on the
`non_pregnant_hemoglobin.exposure` accessor added there.

### Changes and notes

**`4274697` — redraw all living simulants at 6w-9m.** `_apply_6w_9m_shifts`
selected simulants by pregnancy outcome (`live_birth` / `stillbirth`), so
partial-term pregnancies were never redrawn from the non-pregnant
distribution. They carried a pregnancy-scale hemoglobin into the 6w-9m
window, which `AnemiaYLDsObserver` scores with *non-pregnancy* thresholds
(`[80, 110, 120]` vs `[70, 100, 110]`). Both halves can't be right at once.

The postpartum hemoglobin module scopes that step to "Among surviving
simulants" with no term restriction, so the mask is now mortality-based:
every living simulant is redrawn. That also fixes the mask in the other
direction — full-term simulants who had died were previously redrawn, which
the same sentence excludes.

Hemorrhage shifts stay scoped to live births and stillbirths, per the
[antepartum maternal disorders module](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_mncnh_portfolio/antepartum_maternal_disorders_module/module_document.html):

> Antepartum hemorrhage risk should only be applied to pregnancies that do
> not end in abortion/miscarriage/ectopic pregnancy [...]

PPH is already assigned to full-term births alone, so that scoping only
matters for APH. `_apply_0_6w_shifts` is unchanged: partial-term simulants
still hold their pregnancy value through 0-6w.

**`6714798` — 6w-9m reconstruction check.** Reconstructs
`hemoglobin.exposure` at both postpartum steps from the non-pregnant draw
plus the four hemorrhage shifts, and asserts as named refutations that the
6w-9m value carries neither the baseline IFA deletion nor either iron
effect. A guard on the deletion term keeps those refutations from passing
vacuously. The 0-6w branch splits the iron effect back out of the
reconstruction, because the postpartum modifier clips at 0 *before* the iron
modifiers run — iron lands on top of the clipped value rather than inside
the clip.

**`cf8943a` — point the results tests at `model38.3`** and record the five
regenerated executed results notebooks, mirroring the parent commit
`c89d71b`, so the results-dir pointer and the notebooks that read it move
together.

**`e5076c3` — upper-bound `vivarium_dependencies` and `vivarium_inputs`** to
the pre-monorepo dependency set, alongside the `vivarium_build_utils<=3.3.2`
pin `setup.py` already carries. Both requirements were previously unbounded,
so a fresh env build resolved them forward past the monorepo cutover and
failed. `vivarium_dependencies` 1.1.0+ carries the post-monorepo set, and
`vivarium_inputs` has to stay in the vbu-3.x-compatible window (7.1.x wants
vbu `<3.0.0`; 7.2.10+ and all of 8.x want `>=4.0.0`) — hence
`>=7.2.2,<7.2.10`. These pins are what made the model38.3 environments
buildable, and therefore what made the `--runslow` suite below runnable.

### Known divergence, accepted deliberately

`AntepartumHemorrhage.on_time_step` still assigns APH to every pregnancy
(`maternal_disorders.py:349`), which the module doc above prohibits. This PR
scopes the *hemoglobin shift* rather than the assignment, so partial-term
simulants continue to carry `antepartum_hemorrhage == True` — visible in APH
mortality and in any observer stratified on it.

That is intentional, because narrowing the assignment is **not** a local
change. The artifact's APH incidence risk is normalized by pregnancies, not
births (`_load_hemorrhage_incidence`, `loader.py:2081`, divides by
`POPULATION.SCALING_FACTOR`, where PPH divides by `POPULATION.BIRTH_RATE` at
`loader.py:2084`), while the
[APH cause model doc](https://vivarium-research.readthedocs.io/en/latest/models/causes/maternal_disorders/gbd_2021_mncnh/antepartum_hemorrhage.html)
specifies `ir = (1 - postpartum_fraction) * incidence_c367 / birth_rate`
with a denominator of "all births (still and live births)". The code diverges
on both the denominator and the eligible population, and the two divergences
co-vary — restricting eligibility alone would drop APH case counts by roughly
the partial-term share of pregnancies (~43%) and break the
`aph_count / n_preg` ballpark check in
`interactive_simulation_hemorrhage.ipynb`.

A doc-conformant fix would need the loader denominator, the eligibility
filter, and that notebook's denominator changed together, plus a new artifact
build and re-validation of APH incidence and mortality. Since APH is slated
for removal from the model, that work was judged not worth doing and this
divergence is being accepted rather than fixed.

### Verification and Testing

**Interactive notebook.**
`tests/model_notebooks/interactive/interactive_simulation_hemoglobin.ipynb`,
all four scenarios at 60k population (executed notebook committed):

```
baseline: 6w-9m hemoglobin reconstructs from the non-pregnant draw (max abs
err 0.000e+00); carries no IFA deletion (would shift it by -4.5788 g/L) and
no iron effect (oral IFA +9.9223, IV iron +21.6597 g/L). 26079 partial-term
living simulants redrawn unshifted and 33803 full-term redrawn and shifted;
118 dead held at their pregnancy value; 0 clipped at 0 by the shifts.
```

0-6w reconstructs to 1.42e-14, 6w-9m to exactly 0.0, in every scenario.
`All final YLD checks passed` — the anemia person-time and YLD asserts
compare against `sim.get_results()`, so the observer agrees with the new
behavior and not only the reconstruction.

**V&V run: `model38.3`** (`results/model38.3/*`, Jobmon 612816–612818). A
paired rerun of `model38.2` — same seven Standard V&V scenarios, same pinned
draws and seeds, same 20k population, same `aph/` artifact set and launch
environment — with this branch's two behavioral commits as the only delta, so
late-postpartum anemia YLDs diff against 38.2 directly.

Baseline scenario, partial-term simulants at `late_postpartum` as a share of
*all* anemia YLDs:

| | Ethiopia | Nigeria | Pakistan |
|---|---|---|---|
| model38.2 | 38.03% | 38.23% | 8.59% |
| model38.3 | 19.86% | 20.98% | 3.68% |
| total baseline anemia YLDs | −22.7% | −21.8% | −5.1% |

Pakistan moves least because its partial-term share of pregnancies is much
lower. **This is results-changing** — that was the intent: the pre-fix share
was an artifact of scoring pregnancy-scale hemoglobin against non-pregnancy
thresholds. The `is_alive` correction is results-neutral, as
`AnemiaYLDsObserver` already filters on `is_alive`.

**Test suite** — `pytest --runslow` in both environments, nothing excluded:

| | result |
|---|---|
| *vivarium_gates_mncnh_simulation*, `--runslow` | 7 passed, 8 skipped |
| *vivarium_gates_mncnh_artifact*, `--runslow -k "not results"` | 12 passed |
| *vivarium_gates_mncnh_artifact*, `--runslow -k "results"` | 5 passed (33:24) |

The results notebooks resolve `MODEL_RESULTS_DIR = model38.3` and regenerated
with 0 error outputs, confirming they read the new run rather than stale 38.2
data. `black` / `isort` clean.

- [x] model results directory is up to date (`model38.3`)

### Next steps

- [ ] Merge this pull request


tests/model_notebooks/interactive/executed/interactive_simulation_hemoglobin.ipynb
```diff
@@ -2,13 +2,13 @@
  "cells": [
```
zmbc, line 2: <p>I didn't realize that you were going to write this check, feel free to pass it off here (probably to Tyler) to address the below comments.</p><p> </p><p>"That period represents the weeks after a birth they did not have" is not correct. It represents the actual weeks after their pregnancy ended. It's just that they have never experienced APH or PPH, so they aren't shifted by it.</p><p> </p><p>If this branch currently does not include the restriction of APH to full-term pregnancies, then the check should not be scoped in this way, and the shift should be applied even to partial-term pregnancies.</p><p> </p><p>The stuff about alive is a bit confusing, surely we should just be excluding those people from all checks?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/369/discussion/#comment-3776921569'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='73'/>

zmbc, line 2: <pre class="ql-syntax" spellcheck="false">Line #3.    pph_shift_0_6w = art.load('risk_effect.hemorrhage_hemoglobin_shift.pph_shift_0_6w')[draw].item()
</pre><p>TIL <code>.item()</code>  ! I need this all the time!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/369/discussion/#comment-3776921579'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='74'/>

zmbc, line 2: <pre class="ql-syntax" spellcheck="false">Line #19.    assert ifa_deletion &gt; 1, "IFA deletion too small for the 6w-9m checks below to discriminate"
</pre><p>I don't understand this, but it should probably be removed anyway if the unreachable code below is removed</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/369/discussion/#comment-3776921589'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='74'/>

zmbc, line 2: <pre class="ql-syntax" spellcheck="false">Line #92.            if step_name == 'late_postpartum':
</pre><p>I think this block is not useful, since the assert above would always fail if one of these would fail. I would favor removing it</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/ihmeuw/vivarium_gates_mncnh/pull/369/discussion/#comment-3776921599'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='75'/>

zmbc (approved):
Overall results look good, I would like to improve the check a little bit but I am pretty confident that the simulation is correct, except maybe in not applying APH/PPH shifts to partial term (which is really more of a merge conflict with the temporal ordering updates branch than anything)

setup.py
```diff
@@ -42,7 +42,9 @@
         long_description = f.read()
 
     install_requirements = [
-        "vivarium_dependencies<2.0.0",
+        # Pin to <1.1.0 for the same reason as vivarium_build_utils below: 1.1.0+ carries the
+        # post-monorepo dependency set, while this repo is still on the standalone names.
```
patricktnast, line 46: > this repo is still on the standalone names

This is not supposed to be true, right?

hussain-jafari: pre monorepo epic :)

